PR #10, opened 2019-08-21
Vat checks, documentation, organization

ehildenb:


mkr-mcd.md
```diff
@@ -148,9 +148,54 @@ This allows us to enforce properties after each step, and restore the old state
...
+    syntax VatStep ::= "invariant"
+ // ------------------------------
+    rule <k> Vat . invariant => Vat . exception ... </k> [owise]
+    rule <k> Vat . invariant => .               ... </k>
+         <debt> DEBT:Int </debt>
+         <Line> LINE:Int </Line>
+         <vice> VICE:Int </vice>
+         <dai>  DAI      </dai>
+         <sin>  SIN      </sin>
+      requires DEBT >=Int 0 andBool DEBT <=Int LINE
```
ehildenb: More accurate to say "No change to Debt would cause DEBT > LINE, still possible for LINE to change and make DEBT > LINE". Better to compare to previous Debt, and only enforce this check on new debt if there is an increase in the debt (but no change to line).

ehildenb: Only authorized Vat calls (eg fold) are allowed to violate the LINE/DEBT relationship. Note though, that `drip` from `Jug` is non-auth, but it calls `fold` in an auth manner.

mkr-mcd.md
```diff
@@ -181,62 +226,152 @@ This allows us to enforce properties after each step, and restore the old state
...
+                      </vat>-fragment
+                    )
+           ...
+         </vatStack>
+         <vat>
+           <ilks> ...   ILKID          |-> ILK ... </ilks>
+           <urns> ... { ILKID , ADDR } |-> URN ... </urns>
+           ...
+         </vat>
+      requires urnBalance(ILK, URN) <=Int urnBalance(ILK', URN')
```
ehildenb: "less risky" - anyone can interact with a CDP if it's to pay debt or add collateral.

PR #21, opened 2019-09-03
Stubs - Modular Model and documentation of remaining work

ehildenb:


stabilize.md
```diff
@@ -0,0 +1,180 @@
...
+    syntax VowStep ::= "flog" Int
+ // -----------------------------
+
+    syntax VowStep ::= "heal" Rad
+ // -----------------------------
+
+    syntax VowStep ::= "kiss" Rad
+ // -----------------------------
+
+ // TODO: "flop" and "flap" seem to conflict with the config cell names
```
gtrepta: There's either a way around this in K or we need to rename something. I say renaming the configuration cells is preferred since then the user can use Vow.flop/Vow.flap as expected.

PR #39, opened 2019-09-16
Semantics of Flip

dwightguth:
This will not kompile until the necessary changes to K are merged and the submodule is bumped, but I wanted to solicit feedback early.

### DIFF
--- a/cdp-core.md
+++ b/cdp-core.md
@@ -113,7 +113,7 @@ Vat CDP State
         <cat>
           <cat-ward> .Map </cat-ward>
           <cat-ilks> .Map </cat-ilks>
-          <cat-live> 0    </cat-live>
+          <cat-live> 1    </cat-live>
         </cat>
         <spotStack> .List </spotStack>
         <spot>
--- a/cdp-core.md
+++ b/cdp-core.md
@@ -198,8 +198,8 @@ By adjusting the `<vat-ward>`, you can upgrade contracts in place by deploying a
 **TODO**: `rely` and `deny` should be `note`.
 
 ```k
-    syntax VatStep ::= AuthStep
- // ---------------------------
+    syntax VatAuthStep ::= AuthStep
+ // -------------------------------
     rule <k> Vat . auth => . ... </k>
          <msg-sender> MSGSENDER </msg-sender>
          <vat-ward> ... MSGSENDER |-> true ... </vat-ward>
--- a/collateral.md
+++ b/collateral.md
@@ -13,19 +13,31 @@ module COLLATERAL
 
     configuration
       <collateral>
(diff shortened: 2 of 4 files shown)


dwightguth:
The reason I wrote the can* functions like that was so that the more complicated ones would have a one to one correspondence between requires statements and rules, thus not requiring us to make one big rule with a whole bunch of cases. I think having one huge boolean expression is less readable than this, but I'm open to other suggestions.

gtrepta:
> The reason I wrote the can* functions like that was so that the more complicated ones would have a one to one correspondence between requires statements and rules, thus not requiring us to make one big rule with a whole bunch of cases. I think having one huge boolean expression is less readable than this, but I'm open to other suggestions.

I wrote [each requirement on its own line](https://github.com/runtimeverification/mkr-mcd-spec/pull/37/commits/9c8d49f3bab78060c2233bd8c247a9c7e3f9d78d#diff-5f62d05d7c858e7c90b6ad7266578024R131) in the rule and thought it looked ok.

kmcd.md
```diff
@@ -51,6 +51,8 @@ module KMCD
           <end-out>  .Map </end-out>  // mapping (bytes32 => mapping (address => uint256)) Int     |-> Wad
         </end>
       </kmcd>
+
+    syntax Dummy ::= "end"
```
ehildenb: what's going on here?

collateral.md
```diff
@@ -47,21 +56,192 @@ module COLLATERAL
...
+                             tic: 0,
+                             end: NOW +Int TAU,
+                             usr: USR,
+                             gal: GAL,
+                             tab: TAB)
+         ...</flip-bids>
 
     syntax FlipStep ::= "tick" Int
  // ------------------------------
+    rule <k> Flip ILK . tick ID => Flip ILK . exception ... </k>
```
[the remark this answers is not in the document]
ehildenb: I agree, it would be nice to ensure the semantics is total via `owise` rules, and only define the positive cases if possible.

If it becomes really difficult to define the positive cases, we can re-evaluate this decision.

cdp-core.md
```diff
@@ -113,7 +113,7 @@ Vat CDP State
         <cat>
           <cat-ward> .Map </cat-ward>
           <cat-ilks> .Map </cat-ilks>
-          <cat-live> 0    </cat-live>
+          <cat-live> 1    </cat-live>
```
ehildenb: Can we use `true` instead?

gtrepta: Let's make that into a separate PR.

gtrepta: I made a card for it.

https://github.com/runtimeverification/mkr-mcd-spec/projects/1#card-26514903

Issue #24, opened 2019-09-03
Semantics of wrapped Eth


ehildenb:
Similar to Dai, need an "interface" semantics for eth, should be enough to just model `transfer` event.

ehildenb:
We've decided that only Dai will get a specific semantics, others will be generic gems.

PR #44, opened 2019-09-17
WETH Semantics

nanexcool:
Copied from `dai.md`, includes:

- transfer
- transferFrom
- approve

### DIFF
--- a/weth.md
+++ b/weth.md
@@ -1,0 +1,119 @@
+WETH Token
+=========
+
+The WETH token represents an ERC20 compatible version of ETH.
+
+```k
+requires "kmcd-driver.k"
+
+module WETH
+    imports KMCD-DRIVER
+
+    configuration
+      <weth>
+        <weth-stack> .List </weth-stack>
+        <weth-state>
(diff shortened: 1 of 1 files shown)


weth.md
```diff
@@ -0,0 +1,119 @@
...
+        <weth-state>
+          <weth-totalSupply> 0    </weth-totalSupply>
+          <weth-account-id>  0    </weth-account-id>
+          <weth-balance>     .Map </weth-balance>     // mapping (address => uint)                      Address |-> Int
+          <weth-allowance>   .Map </weth-allowance>   // mapping (address => mapping (address => uint))
+        </weth-state>
+      </weth>
+
+    syntax AllowanceAddress ::= "{" Address "->" Address "}"
+ // --------------------------------------------------------
```
ehildenb: Hmmm, this will probably be problematic because the same symbol is used in `dai.md`. So I think we want to declare this in `kmcd-driver.md` instead a single time, and use it in both places.

ehildenb:
As per #24 , we have decided that only Dai will get specific semantics, wrapped Eth will just be another generic gem.

Issue #38, opened 2019-09-16
Implement Flop semantics


gtrepta:
The Flop is a debt auction, used so the Vow can pay off its Sin. Bidders bid a fixed amount of DAI for a lot of MKR, but bid for smaller lots on each bid, minimizing the inflation of MKR.

- [x] Flop . kick
Starts an auction
- [x] Flop . tick
Extends the end time of the auction when no one has made a bid
- [x] Flop . dent
Called by a user to make a bid.
- [x] Flop . deal
Settles the auction.
- [x] Flop . cage
Global Settlement. Freezes the auction
- [x] Flop . yank
Global Settlement. Refunds the current bid.

PR #37, opened 2019-09-13
Flap Semantics

gtrepta:
Implementation for #32

PR #46, opened 2019-09-17
Semantics of Flop

gtrepta:
Implements #38 

Needs #37 to merge first.


gtrepta:
Will re-open after #51 and Gem semantics gets merged.

PR #48, opened 2019-09-17
Vat refactoring

dwightguth:


cdp-core.md
```diff
@@ -550,6 +452,14 @@ This is quite permissive, and would allow the account to drain all your locked c
            USERW |-> ( DAIW => DAIW +Int (RATE *Int DART) )
            ...
          </vat-dai>
+         <vat-Line> LINE </vat-Line>
+      requires (DART <=Int 0
+        orBool ((ILKART +Int DART) *Int RATE <=Int ILKLINE andBool DEBT +Int (RATE *Int DART) <=Int LINE))
```
ehildenb: Either format like so:

```
requires (        DART <=Int 0
           orBool (...)
         )
 andBool ...
```

Or put all the `orBool` on one line, to make it clear how the or/and is nested visually without having to count parens.

Issue #47, opened 2019-09-17
Generic `<gem>` semantics


ehildenb:
`GemLike`s are used in many places throughout the solidity files. For examples:

-   They are used by `join.sol` for generic tokens which can transfer between users, mint, and burn.
-   They are used by `flap` and `flop` to interact with the MKR token when making auctions.

We need a semantics of these gems, which provides at least the primitives requires by `join.sol`, `flap.sol`, and `flop.sol`.

PR #52, opened 2019-09-18
Semantics of Gem

gtrepta:
Fixes: #47

### DIFF
--- a/collateral.md
+++ b/collateral.md
@@ -62,6 +62,17 @@ module COLLATERAL
     rule <k> Gem ACCTGEM . (transfer ACCTDST VALUE => transferFrom MSGSENDER ACCTDST VALUE) ... </k>
          <msg-sender> MSGSENDER </msg-sender>
 
+    syntax GemStep ::= "mint" Address Int
+ // -------------------------------------
+    rule <k> Gem ACCTGEM . mint ACCTDST VALUE => . ... </k>
+         <gem>
+           <gem-addr> ACCTGEM </gem-addr>
+           <gem-balances>...
+             ACCTDST |-> ( BALANCE_DST => BALANCE_DST +Int VALUE )
+           ...</gem-balances>
+         </gem>
+      requires ACCTDST =/=Int 0
+
     syntax Bid ::= Bid ( bid: Int, lot: Int, guy: Address, tic: Int, end: Int, usr: Address, gal: Address, tab: Int )
  // -----------------------------------------------------------------------------------------------------------------
 
--- a/collateral.md
+++ b/collateral.md
@@ -73,6 +73,18 @@ module COLLATERAL
          </gem>
       requires ACCTDST =/=Int 0
 
+    syntax GemStep ::= "burn" Address Int
+ // -------------------------------------
+    rule <k> Gem ACCTGEM . burn ACCTSRC VALUE => . ... </k>
+         <gem>
+           <gem-addr> ACCTGEM </gem-addr>
+           <gem-balances>...
+             ACCTSRC |-> ( BALANCE_SRC => BALANCE_SRC -Int VALUE )
+           ...</gem-balances>
+         </gem>
+      requires ACCTSRC =/=Int 0
+       andBool BALANCE_SRC >=Int VALUE
+
     syntax Bid ::= Bid ( bid: Int, lot: Int, guy: Address, tic: Int, end: Int, usr: Address, gal: Address, tab: Int )
  // -----------------------------------------------------------------------------------------------------------------
 
--- a/collateral.md
+++ b/collateral.md
@@ -55,7 +55,8 @@ module COLLATERAL
              ACCTDST |-> ( BALANCE_DST => BALANCE_DST +Int VALUE )
            ...</gem-balances>
          </gem>
-      requires BALANCE_SRC >=Int VALUE
+      requires VALUE >=Int 0
+       andBool BALANCE_SRC >=Int VALUE
 
     syntax GemStep ::= "move" Address Address Int
  // ---------------------------------------------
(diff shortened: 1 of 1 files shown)


collateral.md
```diff
@@ -26,6 +32,63 @@ module COLLATERAL
...
+           <gem-balances>...
+             ACCTSRC |-> ( BALANCE_SRC => BALANCE_SRC -Int VALUE )
+             ACCTDST |-> ( BALANCE_DST => BALANCE_DST +Int VALUE )
+           ...</gem-balances>
+         </gem>
+      requires BALANCE_SRC >=Int VALUE
+
+    syntax GemStep ::= "move" Address Address Int
+ // ---------------------------------------------
+    rule Gem ACCTGEM . (move ACCTSRC ACCTDST VALUE => transferFrom ACCTSRC ACCTDST VALUE)
```
ehildenb:
```suggestion
    rule <k> Gem ACCTGEM . (move ACCTSRC ACCTDST VALUE => transferFrom ACCTSRC ACCTDST VALUE) ... </k>
```

Either wrap in `<k>` tags or make a `macro` rule that looks like this:

```
rule move ACCTSRC ACCTDST VALUE => transferFrom ACCTSRC ACCTDST VALUE [macro]
```

I don't like the implicit completion of `<k> ... </k>` that happens when you don't wrap in `<k>` tags explicitly.

ehildenb: But I would choose to wrap in `<k> ... </k>` tags since the other rules do too.

collateral.md
```diff
@@ -26,6 +32,63 @@ module COLLATERAL
...
+    syntax GemStep ::= "mint" Address Int
+ // -------------------------------------
+    rule <k> Gem ACCTGEM . mint ACCTDST VALUE => . ... </k>
+         <gem>
+           <gem-addr> ACCTGEM </gem-addr>
+           <gem-balances>...
+             ACCTDST |-> ( BALANCE_DST => BALANCE_DST +Int VALUE )
+           ...</gem-balances>
+         </gem>
+      requires ACCTDST =/=Int 0
```
ehildenb: Where is this constraint coming in? I don't see any reason that `ACCTDST` can't be `0`.

Also, we should check that `VALUE >=Int 0`.

collateral.md
```diff
@@ -26,6 +32,63 @@ module COLLATERAL
...
+ // -------------------------------------
+    rule <k> Gem ACCTGEM . burn ACCTSRC VALUE => . ... </k>
+         <gem>
+           <gem-addr> ACCTGEM </gem-addr>
+           <gem-balances>...
+             ACCTSRC |-> ( BALANCE_SRC => BALANCE_SRC -Int VALUE )
+           ...</gem-balances>
+         </gem>
+      requires ACCTSRC =/=Int 0
+       andBool BALANCE_SRC >=Int VALUE
```
ehildenb: Similarly, not sure where `ACCTSRC =/=Int 0` is coming from. And we should check that `VALUE >=Int 0`.

In the solidity code for `join.sol`, for example, we check that the transfered `wad` is greater than 0 https://github.com/makerdao/dss/blob/25110e12b8323a0d1d8f45b81ebf0dbd4d740818/src/join.sol#L71

collateral.md
gtrepta: It came from a `require(sender != address(0))` statement in the ERC20 solidity file I referenced, but now that you mention it I guess it doesn't matter in this model.

Similarly, the ERC20 contract was using `uint` for the value, so it didn't need to make sure it's positive. I'll add in the check.

[The ERC20.sol I referenced](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol)

collateral.md
```diff
@@ -26,6 +32,63 @@ module COLLATERAL
         </flips>
       </collateral>
 
+    syntax MCDContract ::= GemContract
+    syntax GemContract ::= "Gem" Address
+    syntax MCDStep ::= GemContract "." GemStep [klabel(gemStep)]
+ // ------------------------------------------------------------
+    rule contract(Gem ACCTGEM . _) => Gem ACCTGEM
+    rule [[ address(Gem ACCTGEM) => ACCTGEM ]] <gem-addr> ACCTGEM </gem-addr>
```
ehildenb: Is there any reason to have `<gem-addr> ACCTGEM </gem-addr>` in here? Or can we just do:

```
rule address(Gem ACCTGEM) => ACCTGEM
```

Actually, this seems kinda strange to me anyway. In `Flip_`, the extra parameter is the `Ilk` identifier, _not_ the contract address. The second field is the contract address.

So I would almost think we would want a configuration that looks like this:

```
<gem ...>
  <gem-id> "":String </gem-id>
  <gem-addr> 0:Address </gem-addr>
  <gem-balances> .Map </gem-balances>
</gem>

syntax GemContract ::= "Gem" String
rule contract(Gem GEMID . _) => Gem GEMID
rule [[ address(Gem GEMID) => ACCTGEM ]] <gem-id> GEMID </gem-id> <gem-addr> ACCTGEM </gem-addr>
```

So that we can have tokens like `"MKR"` and `"DAI"` and `"WETH"` to identify the gems, but store the account address as a different field.

ehildenb:
We should probably also do `Gem _ . push` and `Gem _ . pull`, which are just macros for `Gem _ . transferFrom`.

collateral.md
dwightguth: I agree that giving  the gem a name instead of identifying it by an integer is probably better. But then, I am also pretty sure the intent of the Ilk identifier of type bytes32 is also that it's supposed to be a string, but we're representing it as an Int...

gtrepta: Well, an `Address` can also be a string, so I thought we could use the named tokens as the addresses.

ehildenb: Well the point would be to separate the Gem identifier and the account address, not to conflate them. Similar to how Flip is identified by it's `Ilk`, but also has a field saying which contract implements it.

PR #57, opened 2019-09-23
Semantics of Pot

### LINKED ISSUE #11: Semantics of `Pot`
### Configuration

- [x] `<pie>`
- [x] `<Pie>`
- [x] `<dsr>`
- [x] `<chi>`
- [x] `<vow>`
- [x] `<rho>`

### Operations

- [ ] `pot . drip`
- [ ] `pot . join`
- [ ] `pot . exit`

daejunpark:
fix #11

### DIFF
--- a/rates.md
+++ b/rates.md
@@ -16,9 +16,10 @@ module RATES
           <pot-pies> .Map      </pot-pies> // mapping (address => uint256) Address |-> Wad
           <pot-pie>  0:Wad     </pot-pie>
           <pot-dsr>  0:Ray     </pot-dsr>
-          <pot-chi>  0:Ray     </pot-chi>
+          <pot-chi>  0:Rat     </pot-chi> // arbitrary precision
           <pot-vow>  0:Address </pot-vow>
           <pot-rho>  0         </pot-rho>
+          <pot-live> false     </pot-live>
         </pot>
       </rates>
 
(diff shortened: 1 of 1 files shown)


rates.md
```diff
@@ -36,23 +37,40 @@ module RATES
...
          <pot-dsr> DSR </pot-dsr>
          <pot-vow> VOW </pot-vow>
-      requires TIME >=Int RHO
+         <pot-pie> PIE </pot-pie>
+      requires TIME >=Rat RHO
+       andBool DSR >=Rat 1
 
     syntax PotStep ::= "join" Wad
  // -----------------------------
+    rule <k> Pot . join WAD => call Vat . move MSGSENDER THIS ( CHI *Rat WAD ) ... </k>
```
gtrepta: Looks like the `<pot-chi>` cell is needed in this rule.

rates.md
```diff
@@ -36,23 +37,40 @@ module RATES
...
          <this> THIS </this>
          <currentTime> TIME </currentTime>
-         <pot-chi> CHI => CHI +Int (((#pow(DSR, TIME -Int RHO) *Int CHI) -Int CHI) ) </pot-chi>
+         <pot-chi> CHI => CHI *Rat DSR ^Rat (TIME -Rat RHO) </pot-chi>
          <pot-rho> RHO => TIME </pot-rho>
          <pot-dsr> DSR </pot-dsr>
          <pot-vow> VOW </pot-vow>
-      requires TIME >=Int RHO
+         <pot-pie> PIE </pot-pie>
+      requires TIME >=Rat RHO
```
gtrepta: We're strictly using Int for time values. I guess this is harmless but probably better to stick to Int for clarity.

rates.md
```diff
@@ -36,23 +37,40 @@ module RATES
...
+         <this> THIS </this>
+         <currentTime> TIME </currentTime>
+         <msg-sender> MSGSENDER </msg-sender>
+         <pot-pies> ... MSGSENDER |-> ( MSGSENDER_PIE => MSGSENDER_PIE +Rat WAD ) ... </pot-pies>
+         <pot-pie> PIE => PIE +Rat WAD </pot-pie>
+      requires TIME ==Rat RHO
 
     syntax PotStep ::= "exit" Wad
  // -----------------------------
+    rule <k> Pot . exit WAD => call Vat . move THIS MSGSENDER ( CHI *Rat WAD ) ... </k>
```
gtrepta: Need `<pot-chi>` here as well.

kmcd-driver.md
```diff
@@ -155,14 +155,17 @@ Base Data
...
- // ------------------
+    syntax Ray
+    syntax Ray = Rat
+ // ----------------
 
-    syntax Rad ::= Int
- // ------------------
+    syntax Rad
+    syntax Rad = Rat
+ // ----------------
```
ehildenb: Why is this happenning? We don't need to declare `syntax Rad` ahead of time I don't think?

rates.md
```diff
@@ -13,12 +13,13 @@ module RATES
...
-          <pot-pie>  0         </pot-pie>
-          <pot-dsr>  0         </pot-dsr>
-          <pot-chi>  0         </pot-chi>
+          <pot-pies> .Map      </pot-pies> // mapping (address => uint256) Address |-> Wad
+          <pot-pie>  0:Wad     </pot-pie>
+          <pot-dsr>  0:Ray     </pot-dsr>
+          <pot-chi>  0:Rat     </pot-chi> // arbitrary precision
           <pot-vow>  0:Address </pot-vow>
           <pot-rho>  0         </pot-rho>
+          <pot-live> 0         </pot-live>
```
ehildenb: This should be a `Bool` since it's semantically a boolean.

kmcd-driver.md
daejunpark: there was a bug in the sort synonym implementation. seems it is fixed now https://github.com/kframework/k/commit/da5cd7077914dd5c02b7431dc43f2cdfb6b7a918

rates.md
```diff
@@ -16,9 +16,10 @@ module RATES
           <pot-pies> .Map      </pot-pies> // mapping (address => uint256) Address |-> Wad
           <pot-pie>  0:Wad     </pot-pie>
           <pot-dsr>  0:Ray     </pot-dsr>
-          <pot-chi>  0:Ray     </pot-chi>
+          <pot-chi>  0:Rat     </pot-chi> // arbitrary precision
           <pot-vow>  0:Address </pot-vow>
           <pot-rho>  0         </pot-rho>
+          <pot-live> false     </pot-live>
```
ehildenb: The `constructor` for `Pot` starts it with `1`, so this should default to `true`.

ehildenb:
```suggestion
          <pot-live> true      </pot-live>
```

daejunpark: @ehildenb do you want others to match the constructor? if so, `dsr` and `chi` should also default to`1`.

ehildenb: Yes please.

rates.md
```diff
@@ -36,23 +37,43 @@ module RATES
...
          <this> THIS </this>
          <currentTime> TIME </currentTime>
-         <pot-chi> CHI => CHI +Rat (((DSR ^Rat (TIME -Int RHO) *Rat CHI) -Rat CHI) ) </pot-chi>
+         <pot-chi> CHI => CHI *Rat DSR ^Rat (TIME -Int RHO) </pot-chi>
          <pot-rho> RHO => TIME </pot-rho>
          <pot-dsr> DSR </pot-dsr>
          <pot-vow> VOW </pot-vow>
+         <pot-pie> PIE </pot-pie>
       requires TIME >=Int RHO
+       andBool DSR >=Rat 1 // to ensure positive interest rate
```
ehildenb: Use `NOW` instead of `TIME` as a variable name, for uniformity with the rest of the K code.

Also, where did this check `DSR >=Rat 1` come from? I don't think it's guaranteed that interest rates are always giving the user money.

rates.md
```diff
@@ -36,23 +37,43 @@ module RATES
...
  // -----------------------------
+    rule <k> Pot . join WAD => call Vat . move MSGSENDER THIS ( CHI *Rat WAD ) ... </k>
+         <this> THIS </this>
+         <currentTime> TIME </currentTime>
+         <msg-sender> MSGSENDER </msg-sender>
+         <pot-pies> ... MSGSENDER |-> ( MSGSENDER_PIE => MSGSENDER_PIE +Rat WAD ) ... </pot-pies>
+         <pot-pie> PIE => PIE +Rat WAD </pot-pie>
+         <pot-chi> CHI </pot-chi>
+         <pot-rho> RHO </pot-rho>
+      requires TIME ==Int RHO
```
ehildenb: TIME => NOW

rates.md
daejunpark: https://github.com/makerdao/dss/blob/master/src/pot.sol#L139
`sub` reverts when the first argument (`rmul(rpow(dsr, now - rho, ONE), chi)`) is less than the second (`chi`), which means that it reverts when `rpow(dsr, now - rho, ONE) < ONE`, i.e., `dsr ^ (now - rho) < 1`, which in turn means `dsr < 1` since `now >= rho`.

daejunpark: If it is not the intended business logic, then the code is wrong.

PR #58, opened 2019-09-24
Convert from Int to Rat where appropriate

dwightguth:


collateral.md
```diff
@@ -231,7 +231,7 @@ module COLLATERAL
          <flip-bids>...
            ID |-> Bid(... bid: BID, lot: LOT, guy: GUY, tab: TAB) => .Map
          ...</flip-bids>
-      requires GUY =/=Int 0 andBool BID <Int TAB
+      requires GUY =/=K 0 andBool BID <Rat TAB
```
ehildenb: Hmmmm, why `=/=K` here?

ehildenb: Ah nvm, I see that `GUY` is an `Address` now.

PR #61, opened 2019-09-24
semantics of join

dwightguth:


collateral.md
```diff
@@ -233,5 +240,52 @@ module COLLATERAL
...
+    rule [[ address(GemJoin GEMID) => ACCTJOIN ]] <gemJoin-gem> GEMID </gemJoin-gem> <gemJoin-addr> ACCTJOIN </gemJoin-addr>
+    rule <k> GemJoin _ . _ => exception ... </k> [owise]
+
+    syntax GemJoinStep ::= "join" Address Wad
+ // -----------------------------------------
+    rule <k> GemJoin GEMID . join USR AMOUNT
+          => call Vat . slip GEMID USR AMOUNT
+          ~> call Gem GEMID . transferFrom MSGSENDER THIS AMOUNT ... </k>
+         <msg-sender> MSGSENDER </msg-sender>
+         <this> THIS </this>
```
ehildenb: I think we should check that `AMOUNT` is positive. The solidity code effectively does, and it's not clear to me that a negative AMOUNT here makes much sense (esp. since any sensible implementation of `transferFrom` wouldn't allow it).

collateral.md
```diff
@@ -233,5 +240,52 @@ module COLLATERAL
...
+
+    syntax GemJoinStep ::= "exit" Address Wad
+ // -----------------------------------------
+    rule <k> GemJoin GEMID . exit USR AMOUNT
+          => call Vat . slip GEMID MSGSENDER (0 -Rat AMOUNT)
+          ~> call Gem GEMID . transfer USR AMOUNT ... </k>
+         <msg-sender> MSGSENDER </msg-sender>
```
ehildenb: Same here, `AMOUNT >=Int 0`

Issue #54, opened 2019-09-19
Semantics of `Cat`


ehildenb:

gtrepta:
`Cat . bite` makes a call to `Vow . fess`, so the `CDP-CORE` module would need to import `SYSTEM-STABILIZER`. However, `SYSTEM-STABILIZER` is already importing `CDP-CORE`, which would cause a circular dependency.

Also, the Vow semantics were written a while before we made many changes such as implementing Rat, parameterized productions, etc. Cat makes a call to Vat as well as use some of its data. It may be a good idea to go through Vat and get it up to date before we implement Cat, otherwise we would have to also update Cat when we update Vat.

dwightguth:
I think the correct solution to the circular dependency is to move each contract into its own file.

ehildenb:
I agree, let's do one file per contract instead of trying to group them.

PR #72, opened 2019-09-30
Add Cat Semantics

gtrepta:
Fixes: #54

### DIFF
--- a/cat.md
+++ b/cat.md
@@ -1,13 +1,19 @@
 ```k
 requires "kmcd-driver.k"
+requires "flip.k"
+requires "vat.k"
+requires "vow.k"
 
 module CAT
     imports KMCD-DRIVER
+    imports FLIP
+    imports VAT
+    imports VOW
 ```
 
 ```k
-    syntax CatIlk ::= Ilk ( Art: Wad, rate: Ray, spot: Ray, line: Rad ) [klabel(#CatIlk), symbol]
- // ---------------------------------------------------------------------------------------------
+    syntax CatIlk ::= Ilk ( flip: Address, chop: Ray, lump: Wad ) [klabel(#CatIlk), symbol]
+ // ---------------------------------------------------------------------------------------
 ```
 
 Cat Configuration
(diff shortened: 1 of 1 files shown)


cat.md
```diff
@@ -40,9 +46,30 @@ Cat Semantics
...
+    rule <k> Cat . bite ILK URN
+          => call Vat . grab ILK URN THIS VOWADDR (-1 *Rat minRat(INK,LUMP)) (-1 *Rat minRat(ART, minRat(INK,LUMP) *Rat ART /Rat INK))
+          ~> call Vow . fess (ART *Rat RATE)
+          ~> call Flip FLIP . kick URN VOWADDR (ART *Rat RATE *Rat CHOP) minRat(INK,LUMP) 0
+         ...
+         </k>
+         <this> THIS </this>
+         <cat-live> true </cat-live>
+         <cat-ilks>...
+           ID |-> Ilk(... flip: FLIP, chop: CHOP, lump: LUMP)
```
ehildenb: Should this be `ILK |-> ...`? `ID` isn't bound anywhere, so you're grabbing a random one.

cat.md
```diff
@@ -40,9 +46,30 @@ Cat Semantics
...
+         ...</cat-ilks>
+         <vat-ilks>...
+           ID |-> Ilk(... rate: RATE, spot: SPOT)
```
ehildenb: same

cat.md
```diff
@@ -40,9 +46,30 @@ Cat Semantics
...
+         ...</vat-ilks>
+         <vat-urns>...
+           ID |-> Urn( INK, ART )
```
ehildenb: same

ehildenb: Actually, this should be `{ ILK , URN } |-> ...` as the key.

cat.md
```diff
@@ -40,9 +46,30 @@ Cat Semantics
...
+         ...</vat-urns>
+         <vow-addr> VOWADDR </vow-addr>
```
ehildenb: Looks like `vow` is a `file`able field, so we should have a cell `<cat-vow>` which points to the correct `VOWADDR` to use: https://github.com/makerdao/dss/blob/master/src/cat.sol#L101

PR #79, opened 2019-10-18
Transact corrections, udpates to build harness

ehildenb:
-   Enable `Debug` build by default for speed.
-   Strip out erroneous labels added with haskell backend because of K-TERM hack in definition manually.
-   `transact` should have a source address, which is what `msg-sender` and `this` is set to initially.
-   `call` only updates `this`, but `msg-sender` remains unchanged.
-   Single global `owise` rule for throwing exceptions instead of per-file rules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ MAIN_MODULE    := KMCD
 SYNTAX_MODULE  := $(MAIN_MODULE)
 MAIN_DEFN_FILE := kmcd
 
-KOMPILE_OPTS      ?= --emit-json
+KOMPILE_OPTS      :=
 LLVM_KOMPILE_OPTS := $(KOMPILE_OPTS) -ccopt -O2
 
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -22,6 +22,8 @@ K_BIN     := $(K_RELEASE)/bin
 K_LIB     := $(K_RELEASE)/lib
 export K_RELEASE
 
+K_BUILD_TYPE := Debug
```
dwightguth: If your primary goal is for kompile to be as fast as possible, this should be FastBuild, not Debug.

kmcd-driver.md
```diff
@@ -52,47 +52,48 @@ Function Calls
     syntax MCDStep ::= "call" MCDStep
  // ---------------------------------
     rule <k> call AS:AuthStep ~> CONT => contract(AS) . auth ~> AS </k>
```
dwightguth: This is wrong. Msg.sender IS updated by a contract call.

ehildenb: reverted

Makefile
ehildenb: Switched

PR #82, opened 2019-10-25
Attack test: lucash-pot-end

ehildenb:
Fixes: #84 

-   Several bug-fixes discovered while implementing lucash-pot-end attack.
-   Add state initializers for setting up for attack sequence.
-   Add lucash-pot-end attack and expected output, confirmed locally that without the associated fix the attack goes through.
-   Setup testing harness for KMCD simulations.

Note that in the test expected output, the `Pot . file dsr` step does _not_ show up in the `<events>` output, because that step gets rolled back (because of the extra check against `live`, which is set to `false` by the added `Pot . cage`). Additionally, in `<vat-dai>` we can see that `Alice` and `Bobby` end up with the same amount of dai.

### DIFF
--- a/kmcd-driver.md
+++ b/kmcd-driver.md
@@ -198,9 +198,12 @@ We simulate that here.
 
 ```k
     syntax MCDStep ::= "TimeStep"
- // -----------------------------
-    rule <k> TimeStep => . ... </k>
-         <current-time> TIME => TIME +Int 1 second </current-time>
+                     | "TimeStep" Int
+ // ---------------------------------
+    rule <k> TimeStep => TimeStep 1 ... </k>
+
+    rule <k> TimeStep N => . ... </k>
+         <current-time> TIME => TIME +Int N </current-time>
 
     syntax priorities timeUnit > _+Int_ _-Int_ _*Int_ _/Int_
  // --------------------------------------------------------
--- a/kmcd-driver.md
+++ b/kmcd-driver.md
@@ -204,6 +204,7 @@ We simulate that here.
 
     rule <k> TimeStep N => . ... </k>
          <current-time> TIME => TIME +Int N </current-time>
+      requires N >Int 0
 
     syntax priorities timeUnit > _+Int_ _-Int_ _*Int_ _/Int_
  // --------------------------------------------------------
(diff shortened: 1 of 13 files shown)


tests/attacks/lucash-pot-end.mcd
```diff
@@ -0,0 +1,93 @@
...
+transact ADMIN Flap . rely Vow
+transact ADMIN Flop . rely Vow
+
+// Initialize Vat accounts for Vow/Pot/Flap/End
+transact ADMIN Vat . initUser Vow
+transact ADMIN Vat . initUser Pot
+transact ADMIN Vat . initUser Flap
+transact ADMIN Vat . initUser End
+
+// File Vow contract for Pot (since Pot doesn't depend on Vow?)
```
iamchrissmith: This is a `file` because the `Pot` needs to know where the `Vow` is so it can pass it through to the `Vat` when performing `drip`.

ehildenb: ah makes sense

kmcd-driver.md
```diff
@@ -198,9 +198,12 @@ We simulate that here.
...
     syntax MCDStep ::= "TimeStep"
- // -----------------------------
-    rule <k> TimeStep => . ... </k>
-         <current-time> TIME => TIME +Int 1 second </current-time>
+                     | "TimeStep" Int
+ // ---------------------------------
+    rule <k> TimeStep => TimeStep 1 ... </k>
+
+    rule <k> TimeStep N => . ... </k>
+         <current-time> TIME => TIME +Int N </current-time>
```
malturki: It might be a good idea to add the condition that N is positive to eliminate the possibility of having behaviors where time is reversed or behaviors with vacuous time steps (this could be especially important for random testing).

pot.md
```diff
@@ -67,13 +67,32 @@ These parameters are controlled by governance:
...
 ```
 
 **TODO**: Need to use `vow-file` as name to avoid conflict with `<vow>` cell.
 
+Pot Initialization
+------------------
+
+Because data isn't explicitely initialized to 0 in KMCD, we need explicit initializers for various pieces of data.
+
+-   `initUser`: Add the given users accounts to the pies.
```
malturki:
```suggestion
-   `initUser`: Add the given user's account to the pies.
```

kmcd-driver.md
ehildenb: Done.

Issue #83, opened 2019-10-25
`lucash-pot` bug test case


ehildenb:
From the sc-bug-bounties repository, add a KMCD test for the `lucash-pot` example.

PR #89, opened 2019-11-05
Attack-test-lucash-pot

malturki:
Fixes: #83


tests/attacks/lucash-pot.mcd
```diff
@@ -0,0 +1,71 @@
...
+// ---------------
+
+// Create DAI from ETH
+transact "Alice" GemJoin "gold" . join "Alice" 10
+transact "Alice" Vat . frob "gold" "Alice" "Alice" "Alice" 10 10
+
+// Earn Interest
+transact "Alice" Pot . join 10
+transact ADMIN Pot . file dsr 5			// accumulate interest at 5%
+TimeStep 1 day							// for a day
```
ehildenb: Probably only need `1 second` here

tests/attacks/lucash-pot.mcd
```diff
@@ -0,0 +1,71 @@
...
+transact ADMIN Vat . initCDP "gold" "Alice"
+transact "Alice" Vat . hope Pot
+
+// Attack Sequence
+// ---------------
+
+// Create DAI from ETH
+transact "Alice" GemJoin "gold" . join "Alice" 10
+transact "Alice" Vat . frob "gold" "Alice" "Alice" "Alice" 10 10
+
```
ehildenb: We should be `Pot . file dsr 5` then advance clock, then in 1 transaction, `"Alice"` does a `Vat . frob` to draw new Dai, then `Pot . join`, `Pot . drip`, and `Pot . exit`, and `Vat . frob` to exit her Dai and get back the gem + some extra dai for free.

tests/attacks/lucash-pot.mcd
```diff
@@ -0,0 +1,71 @@
...
+// Earn Interest
+transact "Alice" Pot . join 10
+transact ADMIN Pot . file dsr 5			// accumulate interest at 5%
```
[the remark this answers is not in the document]
malturki: I suppose the actual percentage does not matter as long as some positive value is used, but I'll keep the 5 since this is what the attack uses. And will update the comment.

tests/attacks/lucash-pot.mcd
malturki: Again I am trying to stick to what the example says, but you are probably right that the minimum positive amount is enough.

tests/attacks/lucash-pot.mcd
```diff
@@ -0,0 +1,74 @@
...
+// Create DAI from ETH
+transact "Alice" GemJoin "gold" . join "Alice" 10
+transact "Alice" Vat . frob "gold" "Alice" "Alice" "Alice" 10 10
+
+// Earn Interest
+transact ADMIN Pot . file dsr 5            // set interest rate
+transact "Alice" Pot . join 10
+TimeStep 1 second                             // earn interest for a second 
+transact "Alice" Pot . drip
+transact "Alice" Pot . exit 10
```
ehildenb: This is intended behavior. The attack actually is when _no time_ passes between `Pot . join` and `Pot . exit`. So if you move the `TimeStep 1 second` to _before_ the `Pot . join`, then the attack is exhibited.

malturki: done

PR #100, opened 2019-11-09
Properties and violation checking

malturki:
Adds a violation-recording cell and violation checking function, along with four violation properties inspired by the lucash attacks.


kmcd.md
```diff
@@ -80,6 +80,165 @@ State Storage/Revert Semantics
...
+
+    rule unAuthFlapKick(.List) => false
+```
+
+### Earning interest from a pot after End is deactivated (inspired by the lucash-pot-end attack)
+
+The property checks if an `End . cage` is eventually followed by a successful `Pot . join`and `Pot . exit`.
+
+```k
+    syntax Bool ::= potEndInterest(List) [function, functional]
```
ehildenb: After a call to `End . cage`, you don't see a successful call to `Pot . file dsr`

kmcd.md
```diff
@@ -80,6 +80,165 @@ State Storage/Revert Semantics
...
+
+    rule potEndInterestEnd(.List, _) => false
+```
+
+### Earning interest from a pot in zero time (inspired by the lucash-pot attack)
+
+The property checks if a sequence of `Pot . join`, `Pot . drip` and `Pot . exit` is executed in zero time.
+
+```k
+    syntax Bool ::= zeroTimePotInterest(List) [function, functional]
```
ehildenb: A call to `Pot . join` which is preceded by a `TimeStep` more recently than a `Pot . drip` should not succeed.

1.  Check for calls to Pot . drip, if so go to 1. Check for calls to TimeStep, if so go to 2.
2.  Check for calls to Pot . join, if so violation. Check for calls to Pot . drip, if so go to 1.

kmcd.md
malturki: Done

kmcd.md
malturki: Done

kmcd.md
```diff
@@ -72,6 +72,145 @@ State Storage/Revert Semantics
...
+    rule potEndInterestEnd(.List) => false
+```
+
+### Earning interest from a pot in zero time (inspired by the lucash-pot attack)
+
+The property checks if a successful `Pot . join` is preceded by a `TimeStep` more recently than a `Pot . drip'. 
+
+```k
+    syntax Bool ::= zeroTimePotInterest(List) [function, functional]
+ // ----------------------------------------------------------------
```
ehildenb: This will accept trace `TimeStep ; Pot . join ; Pot . drip`, which is an unacceptable trace.

malturki: Right. This is now fixed.  I don’t think though that we will need to have the `Pot. drip` at the end for the detection pattern. `TimeStep ; Pot . join` is bad enough: once an attacker joins while `now > rho`, the damage is done.

PR #105, opened 2019-11-12
Model updates - formatting and constraints on Vat

ehildenb:
Adds constraints on Vat which come implicitly from:

-   Use of `sub` from `SafeMath`
-   Use of `uint` type in Solidity.

-   Formats several of the files.
-   Moves MKR gem initialization to ATTACK-PRELUDE instead of having directly in attack itself.


tests/attacks/lucash-flap-end.mcd.expected
```diff
@@ -60,6 +60,8 @@
       ListItem ( LogNote ( ADMIN , Gem "gold" . mint "Bobby" 20 ) )
       ListItem ( LogNote ( ADMIN , Gem "MKR" . initUser "Alice" ) )
       ListItem ( LogNote ( ADMIN , Gem "MKR" . initUser "Bobby" ) )
+      ListItem ( Exception ( Gem "MKR" . initUser Flap ) )
```
malturki: `transact ADMIN Gem "MKR" . initUser Flap` is generating an exception. Is this intended?

PR #106, opened 2019-11-13
Initial trace generator infrastructure

ehildenb:
Adds the infrastructure for doing trace generation to `kmcd-prelude.md`.

Will move to a separate module in a later PR.


Makefile
```diff
@@ -135,15 +135,18 @@ $(java_kompiled): $(java_files)
 # Test
 # ----
 
+KMCD_GENDEPTH   := 20
+KMCD_RANDOMSEED := 0
```
malturki: Fixing the seed across runs means that the sequence of random numbers generated will be the same every time the tests are run, which can be useful when one needs to reproduce results. Another option here is to use the system clock time value (I suppose something like `date +%s`) as the seed, so that a fresh set of executions are generated at every run.

kmcd-prelude.md
```diff
@@ -144,6 +143,134 @@ module KMCD-PRELUDE
...
+          </generators>
+          <end-gen> .GenStep </end-gen>
+        </kmcd-gen>
+      </kmcd-random>
+
+    syntax DepthBound ::= Int | "*"
+                        | decrement ( DepthBound ) [function]
+ // ---------------------------------------------------------
+    rule decrement(*) => *
+    rule decrement(N) => N -Int 1
```
malturki: We might want to add the condition that `N >Int 0`, to disallow improper usage. And perhaps also another rule `rule decrement(N) => 0 if N <=Int 0` to avoid partiality in this case.

kmcd-prelude.md
```diff
@@ -144,6 +143,134 @@ module KMCD-PRELUDE
...
+    rule randIntBounded(RAND, 0)     => 0
+    rule randIntBounded(RAND, BOUND) => RAND modInt BOUND requires BOUND =/=Int 0
+
+    syntax Rat ::= randRat ( Int ) [function]
+ // -----------------------------------------
+    rule randRat(I) => (I modInt 100) /Rat 100
+
+    syntax Rat ::= randRatBounded ( Int , Rat ) [function]
+ // ------------------------------------------------------
+    rule randRatBounded(I, BOUND) => BOUND *Rat randRat(I)
```
malturki: Is there a reason for using 100 here in `RandRat`? Note that by limiting `randRat` to one hundred possible values here, `randRatBounded` is also limited to a hundred possible values, despite the fact that it's scaled by the bound. My suggestion (Assuming there is no specific reason for picking 100): If the maximum random value that K's `randInt` generates is `RANDMAX`, then define `randRat(I)` as `(I modInt RANDMAX) /Rat (RANDMAX -Int 1). If there is no theoretical maximum, then use any very large number.

kmcd-prelude.md
```diff
@@ -144,6 +143,134 @@ module KMCD-PRELUDE
...
+                     > GenStep ";" GenStep [left]
+ // ---------------------------------------------
+    rule <k> .GenStep => . ... </k>
+
+    rule <k> .GenStep ; GSS => GSS ... </k> [priority(49)]
+    rule <k> GSS ; .GenStep => GSS ... </k> [priority(49)]
+    rule <k> .GenStep | GSS => GSS ... </k> [priority(49)]
+    rule <k> GSS | .GenStep => GSS ... </k> [priority(49)]
+
+    rule <k> GSS DB:DepthBound => #if DB ==K 0 #then . #else (GSS ; (GSS decrement(DB))) | .GenStep #fi ... </k>
```
malturki: Perhaps we might also want to simplify `.GenStep DB => .GenStep`

kmcd-prelude.md
ehildenb: I picked 100 because it's relatively small and I suspect that most attacks aren't going to be much different based on what the chosen random input is.

Makefile
ehildenb: In a later PR I select out a particular seed for each run as a "witness". This is for testing, so we want deterministic runs.

kmcd-prelude.md
ehildenb: Will do.

kmcd-prelude.md
ehildenb: Will do.

PR #108, opened 2019-11-14
Same key rules for map accesses

ehildenb:
-   Formats single-key accesses one line.
-   Adds in many rules for same-key accesses to maps (where no update is expected).


gem.md
```diff
@@ -97,6 +97,15 @@ Gem Semantics
       requires VALUE >=Rat 0
        andBool BALANCE_SRC >=Rat VALUE
```
malturki: In dai.md above, there is also the check that the two accounts must be different, even though we are matching on two entries in the balances map. For consistency, we should probably do the same here: `ACCTSRC =/=K ACCTDST`, right?

gem.md
```diff
@@ -135,11 +140,7 @@ Gem Semantics
...
            <gem-id> GEMID </gem-id>
-           <gem-balances>
-             ...
-             ACCTSRC |-> ( BALANCE_SRC => BALANCE_SRC -Rat VALUE )
-             ...
-           </gem-balances>
+           <gem-balances> ... ACCTSRC |-> ( BALANCE_SRC => BALANCE_SRC -Rat VALUE ) ... </gem-balances>
            ...
          </gem>
       requires VALUE >=Rat 0
```
malturki: Shouldn't we have a check here that the amount being burned is not more than the account's balance (I haven't actually checked the contract source)?

gem.md
ehildenb: Fixed.

gem.md
ehildenb: Fixed.

ehildenb: The `gem` contracts are third-party supplied, so this is just a model of a "generic" gem contract. So it makes sense to make it as "tight" as possible.

PR #109, opened 2019-11-18
Randomly generated witnesses to 4 lucash attacks

ehildenb:
This adds generators for the various steps needed, and witnesses for the 4 lucash attacks.


kmcd-prelude.md
```diff
@@ -275,5 +276,269 @@ module KMCD-GEN
...
+                        | "GenVatMove" Address Address
+                        | "GenVatHope"
+                        | "GenVatHope" Address
+                        | "GenVatHope" Address Address
+ // --------------------------------------------------
+    rule <k> GenVatFrob => GenVatFrob chooseCDPID(I, keys_list(VAT_URNS)) ... </k>
+         <random> I => randInt(I) </random>
+         <vat-urns> VAT_URNS </vat-urns>
+      requires size(VAT_URNS) >Int 0
+
```
malturki: GenVatFrob will fail if the condition is not satisfied (vat urns is empty). I suppose we want the generator to gracefully recover and try some other step in this case.

kmcd-prelude.md
```diff
@@ -275,5 +276,269 @@ module KMCD-GEN
...
+    rule <k> GenVatFrob CDPID => GenVatFrob CDPID ((2 *Rat randRatBounded(I, VAT_GEM)) -Rat VAT_GEM) ... </k>
```
malturki: I think the expression ((2 *Rat randRatBounded(I, VAT_GEM)) -Rat VAT_GEM) means the value will be in the half-open range [- VAT_GEM, VAT_GEM), which I suppose means that the generator will never consider the case when a user allocates all his gem for dai.

kmcd-prelude.md
```diff
@@ -275,5 +276,269 @@ module KMCD-GEN
...
+         <random> I => randInt(I) </random>
+         <vat-gem> ... CDPID |-> VAT_GEM ... </vat-gem>
+
+    rule <k> GenVatFrob { ILKID , ADDRESS } DINK
+          => #fun( DARTBOUND
+                => LogGen ( transact ADDRESS Vat . frob ILKID ADDRESS ADDRESS ADDRESS DINK ((2 *Rat randRatBounded(I, DARTBOUND)) -Rat DARTBOUND) )
```
malturki: similarly here for ((2 *Rat randRatBounded(I, DARTBOUND)) -Rat DARTBOUND)

kmcd-prelude.md
```diff
@@ -275,5 +276,269 @@ module KMCD-GEN
...
+         <vat-ilks>
+           ...
+           ILKID |-> Ilk ( ... rate: RATE, spot: SPOT )
+           ...
+         </vat-ilks>
+         <vat-urns>
+           ...
+           { ILKID , ADDRESS } |-> Urn ( ... ink: URNINK, art: URNART )
+           ...
+         </vat-urns>
```
malturki: GenVatFrob may (in principle) fail here as well if the needed vat-ilks or vat-urns are not found. An additional `owise` rule would do.

ehildenb:
I've made it generate rationals in the closed set, and added hard-coded parameters for various system constants.

Note that we already have the following rule:

```
    rule <k> GS => . ... </k> 
         <events> ... (.List => ListItem(GenStepFailed(GS))) </events> 
      [owise] 
```

PR #112, opened 2019-11-20
Generate random inputs from mcd-pyk script

ehildenb:
This PR fleshes out the `mcd-pyk.py` script to generate the initial configuration with the `ATTACK-PRELUDE` in state, then sets the stage for randomly generating attacks using loaded generators.

Later PR will load actual generators and adjust the tests to use this generator instead of the hand-written ones.

BLOCKED on #113


kmcd-prelude.md
```diff
@@ -233,11 +234,20 @@ module KMCD-GEN
...
          <violation> false </violation>
       requires lengthBytes(BS) >Int 0
+       andBool N >Int 0
+
+    rule <k> GenStep => . ... </k>
+         <generator-next> N </generator-next>
+         <random> BS => tail(BS) </random>
+         <used-random> BS' => BS' +Bytes Int2Bytes(head(BS), 1, BE) </used-random>
+      requires lengthBytes(BS) >Int 0
+       andBool notBool N >Int 0
```
malturki: I'd move `<generator-next> N </generator-next>` two lines below for consistency with other similar rules.

kmcd-prelude.md
```diff
@@ -233,11 +234,20 @@ module KMCD-GEN
  // -----------------------------------
     rule <k> GenStep => GenStepLoad ~> GenStepReplace ... </k>
          <random> BS => tail(BS) </random>
+         <used-random> _ => Int2Bytes(head(BS), 1, BE) </used-random>
```
malturki: Isn't `BS` already a list of bytes? Why `Int2Bytes` in `<used-random>`?

ehildenb: We're appending the `head(BS) :: Int` to the end of `<used-random>` (because it's the next choice we're making). I can possibly use `substrBytes(BS, 0, 1)` instead, so that it remains in the `Bytes` type.

kmcd-prelude.md
ehildenb: Done.

kmcd-prelude.md
ehildenb: I've factored out `headAsBytes` helper function, which now uses `substrBytes(BS, 0, 1)` and makes the sorts more obvious.

PR #126, opened 2019-12-05
update flop spec and tests to account for yank-suck changes

iamchrissmith:
Everett and Musab,

I couldn't find where we had updated the Flop spec to account for the bug fix we did, so I took a shot and making the update myself.  I was able to run:
`make deps` (no changes)
`make build -j12` (got though some missing pieces and then didn't fail)
`make test-execution -j12` (identified missing state, then said no changes from `expected`)

Please let me know if there are other things I should run to ensure this is correct and how I could make any improvements to the changes

### DIFF
--- a/flop.md
+++ b/flop.md
@@ -23,6 +23,7 @@ Flop Configuration
         <flop-pad>    150 /Rat 100 </flop-pad>
         <flop-ttl>    3 hours      </flop-ttl>
         <flop-tau>    2 days       </flop-tau>
+        <flop-vow>    0:Address    </flop-vow>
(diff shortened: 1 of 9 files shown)


flop.md
```diff
@@ -86,6 +87,7 @@ The parameters controlled by governance are:
                       | "ttl" Int
                       | "tau" Int
                       | "pad" Ray
+                      | "vow-file" Address
  // -----------------------------
```
ehildenb:
```suggestion
 // --------------------------------------
```

PR #128, opened 2019-12-10
Measurables on total Vat Dai under user control

ehildenb:


kmcd-props.md
```diff
@@ -14,17 +14,174 @@ module KMCD-PROPS
...
+### Bounded Debt Growth
+
+The Debt growth should be bounded in principle by the interest rates available in the system.
+
+```k
+    syntax Bool ::= totalDebtBounded    ( List             ) [function]
+                  | totalDebtBoundedAux ( List , Rat , Rat ) [function]
+ // -------------------------------------------------------------------
+    rule totalDebtBounded(.List)                           => true
+    rule totalDebtBounded(ListItem(Measure(DEBT, _)) REST) => totalDebtBoundedAux(REST, DEBT, 1) // initial DSR 1
```
ehildenb: Just read the DSR here instead.

ehildenb: Actually we can't do this because the `<pot-dsr>` cell is not accurate at the time we're processing this entry.

PR #130, opened 2019-12-17
Separate and name violations

ehildenb:
Makes it easier to see why a given trace violates some properties by:

-   Turning properties into `ViolationFSM`s instead of having them just be boolean formulas,
-   Having each property being checked be in its own entry in a `Map` with a given name.

### DIFF
--- a/kmcd-props.md
+++ b/kmcd-props.md
@@ -140,10 +140,12 @@ A violation occurs if any of the properties above holds.
 ```k
     syntax Map ::= "#violationFSMs" [function]
  // ------------------------------------------
-    rule #violationFSMs => "Zero-Time Pot Interest Accumulation" |-> zeroTimePotInterest
-                           "Pot Interest Accumulation After End" |-> potEndInterest
-                           "Unauthorized Flip Kick"              |-> unAuthFlipKick
-                           "Unauthorized Flap Kick"              |-> unAuthFlapKick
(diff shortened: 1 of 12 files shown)


kmcd-props.md
```diff
@@ -137,65 +138,150 @@ Violations
...
-                      orBool unAuthFlipKick(EVENTS)
-                      orBool unAuthFlapKick(EVENTS)
-                      orBool potEndInterest(EVENTS)
+    syntax Map ::= "#violationFSMs" [function]
+ // ------------------------------------------
+    rule #violationFSMs => ( "Zero-Time Pot Interest Accumulation" |-> zeroTimePotInterest )
+                           ( "Pot Interest Accumulation After End" |-> potEndInterest      )
+                           ( "Unauthorized Flip Kick"              |-> unAuthFlipKick      )
+                           ( "Unauthorized Flap Kick"              |-> unAuthFlapKick      )
+                           ( "Total Bound on Debt"                 |-> totalDebtBounded(1) )
```
kmbarry1: Why `1` here? Naively, I'd expect zero since the system starts off with no debt.

ehildenb: The `1` here is actually DSR, not `Debt`. So we are assuming we start at 0% interest on DSR.

kmbarry1: Ah, that's right, thanks.

PR #136, opened 2020-01-13
Randomly generate results

ehildenb:
This adds target `test-random`, which can find the attack sequences (with the appropriate breakages to the code).

I've removed one step of pre-determined randomness from each test, and made sure it could still find the relevant attacks. Next I'll try removing more randomness, and see where performance bottlenecks are.

Current runs take about 25s per run, which is way too long. I know we can shave off a decent chunk by skipping the frontend altogether, or using `kserver` to keep a warmed up JVM around. But I think beyond that it will be algorithmic inefficiencies that I'll need to profile.

### DIFF
--- a/mcd-pyk.py
+++ b/mcd-pyk.py
@@ -42,7 +42,7 @@ def kastJSON_haskell(inputJSON, *kastArgs):
     return pyk.kastJSON(MCD_definition_haskell_dir, inputJSON, kastArgs = list(kastArgs))
 
 def krunJSON_llvm(inputJSON, *krunArgs):
-    return pyk.krunJSON(MCD_definition_llvm_dir, inputJSON, krunArgs = list(krunArgs), keepTemp = True)
+    return pyk.krunJSON(MCD_definition_llvm_dir, inputJSON, krunArgs = list(krunArgs))
 
 def krunJSON_haskell(inputJSON, *krunArgs):
     return pyk.krunJSON(MCD_definition_haskell_dir, inputJSON, krunArgs = list(krunArgs))
@@ -104,6 +104,18 @@ def get_init_config(init_term):
     kast_json = { 'format': 'KAST', 'version': 1, 'term': init_term }
     (_, init_config, _) = krunJSON_llvm(kast_json, *randomSeedArgs())
     return pyk.splitConfigFrom(init_config)
+
+def detect_violations(config):
+    (_, configSubst) = pyk.splitConfigFrom(config)
+    properties = configSubst['PROPERTIES_CELL']
+    violations = []
+    def _gatherViolations(fsmMap):
+        if pyk.isKApply(fsmMap) and fsmMap['label'] == '_|->_':
+            if fsmMap['args'][1] == pyk.KConstant('Violated_KMCD-PROPS_ViolationFSM'):
+                violations.append(fsmMap['args'][0]['token'])
+        return fsmMap
+    pyk.traverseTopDown(properties, _gatherViolations)
+    return violations
 
 def steps(step):
     return KApply('STEPS(_)_KMCD-PRELUDE_MCDStep_MCDSteps', [step])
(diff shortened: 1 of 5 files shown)


ehildenb:
Should add, that with one predetermined step removed, it reliably finds all four attacks in 1000s. I haven't tried shorter number of runs, but will begin optimizing performance first.

ehildenb:
Also, it switches the argument order for `mcd-pyk.py` so that multiple seeds can be supplied at start.

ehildenb:
I've checked, and running the backend itself actually only takes 0.03s or so per test, whereas running the frontend is taking 27s. So I'm going to work on skipping the K frontend entirely for my next PR.

tests/attacks/lucash-flap-end.random.mcd.expected
```diff
@@ -39,7 +39,7 @@
             .Map
           </cat-ilks>
           <cat-live>
-            false
+            true
```
kmbarry1: I'm a bit perplexed by this and the other changes to the `live` flags...since this attack specifically requires emergency shutdown to have been activated, I'd think that `false` would be the right thing to have for `live` cell values.

ehildenb: Hmmm, that's true that the attack sequence requires the shutdown to happen to make sense.

I shortened the attack sequences to the exact prefix where the `Violated` shows up in the `<properties>` cell, for the purpose of knowing how long to run the random-tester for. The actual property being checked is that "No non-auth Flap kicks are set off", which is triggered before the shutdown happens (specifically, when the non-authed flap-kick happens).

This arguably isn't a very useful property, so perhaps we should come up with a better property which encompasses this attack and the "No non-auth Flip kicks are set off" attack.

For now, I'll add an item to the meeting notes for us to brainstorm better/more general properties for catching these attacks, and I'll extend these attack sequences again to exhibit the full attack (not just tho non-authed flip/flap kicks).

iamchrissmith: Taking a first stab (in plain English) at the Flip one:
The attack occurs when the end process attempts to reimburse a user for their bid when that user has not submitted a bid before.

So an exception has occurred if there are a sequence of transactions that includes a:
```
yank(uint)  -> vat.move(END, guy, bid)
```
without a preceding:
```
vat.move(guy, gal, bid)
```

iamchrissmith: similarly, we can say the flap bug occurred when there is a:
```
yank(uint) -> gem.move(this, guy, bid)
```

without a preceeding:
```
gem.move(guy, this, bid)
```

ehildenb: Hmmm, these are more general properties, and would catch these particular bugs. I was hoping we could find some measurable we can track though which produces the same bugs, rather than looking for specific sequences of allowed/disallowed actions. Statements about measurables will generalize better to other situations.

ehildenb: I think that the flip attack might be detectable with eth TotalDebtBounded property, or perhasp the Debt == Sum(Dai) property. But I'm not sure.

kmbarry1: The approach Chris lays out will work, but I feel like there should be something more fundamental...something that can be deduced just from looking at the overall dai balances and the bid values of all auctions...i.e. you ought be able to detect it from a snapshot of the various account values post-attack, with no knowledge of the history. I can't quite put my finger on it yet, though...

kmbarry1: One simple idea: if `End.debt  != 0 && End.debt < Vat.debt`, that indicates that debt was generated post-`thaw`, which should not happen, and would detect at least the `flip` attack (I haven't thought about the `flap` attack at all yet).

ehildenb: Well, we already detect if `End.debt` goes up post-`cage` (with `totalDebtBounded` property), so I think we should see if we can get it to find the flip attack using that property.

mcd-pyk.py
```diff
@@ -257,15 +273,37 @@ def addGenerator(generator):
...
+            print()
+            violations = detect_violations(output)
+            if len(violations) > 0:
+                all_violations.append({ 'properties': violations , 'seed': str(curRandSeed), 'output': output })
+    stopTime = time.time()
+
+    elapsedTime = stopTime - startTime
+    perRunTime  = elapsedTime / (numruns * len(randseeds))
+    print('\n\nTime Elapsed: ' + str(elapsedTime))
+    print('\nTime Per Run: ' + str(perRunTime))
```
godsflaw: Just checking that the above codeblock is intention and not just for debugging.

tests/attacks/lucash-flap-end.random.mcd.expected
kmbarry1: On further thought, the property I propose above is indeed too aggressive...e.g. if `skip` gets called post-`thaw` (even on a real auction), `Vat.debt` will increase. This is arguably a flaw in the system itself though!

PR #142, opened 2020-01-15
Generalize generators and speed up execution

ehildenb:
-   Remove argument hints from generators used in Python script (to cover wider swaths of execution).
-   Make `test-random` target generate paths up to 100 bytes deep, and generate 1000 of them.
-   Remove all hints to `test-random` target about which generators to use from `Makefile`.
-   Add `--no-sort-collections` option to `krun` to take us from roughly 4s/run to 2s/run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ test-python-generator: $(execution_tests_random:=.python-out)
 init_random_seeds :=
 
 test-random: mcd-pyk.py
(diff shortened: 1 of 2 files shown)


iamchrissmith:
I removed `automerge` so @kmbarry1 has a chance to review as well if he wants before the bot merges it on my approval

PR #149, opened 2020-01-28
Usability improvements

ehildenb:
-    Remove outdated Vat invariants section (most of the invariants are either not true or subsumed by totalDebtBounded)
-    Make violations record the last non-violating state of the VFSM for easier understanding of what's going on when violations occur.
-    Avoid use of `#ite` in `kmcd-props.md` when it's not needed.


ehildenb:
No, the expected files are based on the model checked into master, which has no violations present in the existing attacks, so `Violated(_)` never shows up anyway.

PR #152, opened 2020-02-02
Add Flap invariants

kmbarry1:
Add two new invariants for the Flap contract. Note that the `flapMkrConsistency` VFSM catches the `lucash-flap-end` attack, so that opens the door to remove the rather narrow `unAuthFlapKick` VFSM.

### DIFF
--- a/flap.md
+++ b/flap.md
@@ -103,15 +103,6 @@ Flap Events
 ```k
     syntax Event ::= FlapKick(Address, Int, Rad, Wad)
  // -------------------------------------------------
-
-    syntax Event ::= FlapTend(Int, Rad, Wad)
- // ----------------------------------------
-
-    syntax Event ::= FlapDeal(Int)
- // ------------------------------
-
-    syntax Event ::= FlapYank(Int)
- // ------------------------------
 ```
 
 Flap Semantics
@@ -169,7 +160,6 @@ Flap Semantics
          <flap-live> true </flap-live>
          <flap-ttl> TTL </flap-ttl>
          <flap-beg> BEG </flap-beg>
-         <frame-events> _ => ListItem(FlapTend(ID, LOT, BID)) </frame-events>
       requires (TIC >Int NOW orBool TIC ==Int 0)
        andBool END  >Int NOW
        andBool LOT ==Rat LOT'
@@ -192,7 +182,6 @@ Flap Semantics
          <current-time> NOW </current-time>
          <flap-bids> ... ID |-> FlapBid(... bid: BID, lot: LOT, guy: GUY, tic: TIC, end: END) => .Map ... </flap-bids>
          <flap-live> true </flap-live>
-         <frame-events> _ => ListItem(FlapDeal(ID)) </frame-events>
       requires TIC =/=Int 0
        andBool (TIC <Int NOW orBool END <Int NOW)
 ```
@@ -219,7 +208,6 @@ Flap Semantics
          <this> THIS </this>
          <flap-bids> ... ID |-> FlapBid(... bid: BID, guy: GUY) => .Map ... </flap-bids>
          <flap-live> false </flap-live>
-         <frame-events> _ => ListItem(FlapYank(ID)) </frame-events>
 ```
 
 ```k
--- a/flap.md
+++ b/flap.md
@@ -103,9 +103,13 @@ Flap Events
 ```k
     syntax Event ::= FlapKick(Address, Int, Rad, Wad)
  // -------------------------------------------------
-```
 
-```k
+    syntax Event ::= FlapTend(Int, Rad, Wad)
+ // ----------------------------------------
+
+    syntax Event ::= FlapDeal(Int)
+ // ------------------------------
+
     syntax Event ::= FlapYank(Int)
  // ------------------------------
 ```
@@ -165,6 +169,7 @@ Flap Semantics
          <flap-live> true </flap-live>
          <flap-ttl> TTL </flap-ttl>
          <flap-beg> BEG </flap-beg>
+         <frame-events> _ => ListItem(FlapTend(ID, LOT, BID)) </frame-events>
       requires (TIC >Int NOW orBool TIC ==Int 0)
        andBool END  >Int NOW
        andBool LOT ==Rat LOT'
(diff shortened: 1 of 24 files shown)


flap.md
```diff
@@ -160,6 +169,7 @@ Flap Semantics
          <flap-live> true </flap-live>
          <flap-ttl> TTL </flap-ttl>
          <flap-beg> BEG </flap-beg>
+         <frame-events> _ => ListItem(FlapTend(ID, LOT, BID)) </frame-events>
```
ehildenb: Do we need these log events? As far as I can tell, you can be equally well served by looking for `LogNote(Flap . tend _ _ _)` in your ViolationFSM below.

kmcd-props.md
```diff
@@ -326,6 +354,50 @@ The property checks if a successful `Pot . join` is preceded by a `TimeStep` mor
...
+```
+
+### Flap Invariants
+
+```k
+    syntax Bool ::= flapDaiGtOrEtSumOfFlapLots() [function, functional]
+ // -------------------------------------------------------------------
+    rule [[ flapDaiGtOrEtSumOfFlapLots() =>
+              #lookup(USERDAI, Flap) >=Rat sumOfAllFlapLots(FLAP_BIDS)
+         ]]
```
ehildenb: I would recommend:

```
    rule [[ flapDaiGtOrEtSumOfFlapLots() => #lookup(USERDAI, Flap) >=Rat sumOfAllFlapLots(FLAP_BIDS) ]]
```

or

```
    rule [[ flapDaiGtOrEtSumOfFlapLots()
         => lookup(USERDAI, Flap) >=Rat sumOfAllFlapLots(FLAP_BIDS)
         ]]
```

For consistency with this and other semantics.

kmcd-props.md
```diff
@@ -326,6 +354,50 @@ The property checks if a successful `Pot . join` is preceded by a `TimeStep` mor
...
 
+### Flap dai consistency
+
+```k
+    syntax ViolationFSM ::= "flapDaiConsistency"
+ // --------------------------------------------
+    rule derive(flapDaiConsistency, FlapKick(_, _, _, _)) => Violated(flapDaiConsistency) requires notBool(flapDaiGtOrEtSumOfFlapLots())
+    rule derive(flapDaiConsistency, FlapTend(_, _, _))    => Violated(flapDaiConsistency) requires notBool(flapDaiGtOrEtSumOfFlapLots())
+    rule derive(flapDaiConsistency, FlapDeal(_))          => Violated(flapDaiConsistency) requires notBool(flapDaiGtOrEtSumOfFlapLots())
+    rule derive(flapDaiConsistency, FlapYank(_))          => Violated(flapDaiConsistency) requires notBool(flapDaiGtOrEtSumOfFlapLots())
```
ehildenb: Instead of matching on all these events, you could just match on the `Measure(_)` event. That way it gets checked every step and not only when `Flap` steps are taken. This is good for 2 reasons:

-   Let's you have a single rule instead of 4.
-   You can then add the given measures you want (`sumOfAllFlapBids` and `sumOfAllFlapLots`) to the `Measure` record, so that this is not being turned into a contextual function.

kmcd-props.md
kmbarry1: Done. Btw what does the `[[ ]]` actually mean here? I cargo-culted it from other code but I'm not sure what it's doing...

ehildenb:
You should back out the update to `deps/k` submodule. Our automated submodule updater will take care of those updates, and it will just cause problems to update it manually.

kmbarry1:
The `deps/k` thing came out of doing a `git rebase`, I don't think there's any diff with `master`...I certainly didn't do it manually.

kmcd-props.md
```diff
@@ -23,17 +23,19 @@ Measurables
...
+         <vat-debt>     DEBT      </vat-debt>
+         <vat-dai>      VAT_DAIS  </vat-dai>
+         <vat-ilks>     VAT_ILKS  </vat-ilks>
+         <vat-urns>     VAT_URNS  </vat-urns>
+         <vat-vice>     VAT_VICE  </vat-vice>
+         <pot-chi>      POT_CHI   </pot-chi>
+         <pot-pie>      POT_PIE   </pot-pie>
+         <end-debt>     END_DEBT  </end-debt>
+         <flap-bids>    FLAP_BIDS </flap-bids>
+         <gem-balances> MKR_BALS  </gem-balances>
```
ehildenb: One problem with this is that the cell `<gem-balances>` shows up inside a `multiplicity="*" type="Map"` cell.

From `gem.md`:

```
    configuration 
      <gems> 
        <gem multiplicity="*" type="Map"> 
          <gem-id>       "":String </gem-id> 
          <gem-wards>    .Set      </gem-wards> 
          <gem-balances> .Map      </gem-balances> // mapping (address => uint256) Address |-> Wad 
        </gem> 
      </gems> 
```

So this will actually just be picking out some _random_ gem to get the balances cell from, not necessarily the `MKR` gem.

You want to say something more like this:

```
<gem> <gem-id> "MKR" </gem-id> <gem-balances> MKR_BALS </gem-balances> ... </gem>
```

This may fix your issue because when you have a `type="Map"` cell, it expects that _every_ rule which mentions any sub-cells of it _also_ mentions the "key" cell, which is the first sub-cell (in this case `<gem-id>` cell). Internally, this is used as an optimization to turn this access into a map lookup instead of a randomized search through all potential cells.

kmbarry1: Doing that at least changed the error message:
```
%map* (%block*)* bitcast (%block* (%block*)* @apply_rule_4865 to %map* (%block*)*)/mnt/stuff/maker/mkr-mcd-spec/deps/k/k-distribution/target/release/k/lib/../bin/llvm-kompile: line 27: 26700 Aborted                 (core dumped) "$(dirname "$0")"/llvm-kompile-codegen "$definition" "$dt_dir"/dt.yaml "$dt_dir" $debug > "$mod"
```

ehildenb:
Hmmm, I can reproduce the issue building, but don't know what's causing it. I'll investigate more tomorrow.

kmcd-props.md
kmbarry1: For the record: was a bug in K, fixed with deps update.

kmcd-props.md
kmbarry1: Done.

flap.md
kmbarry1: Removed them.

kmbarry1:
I think this is ready to go in. There needs to be a follow-up PR to make this detect the `lucash-flap-end` attack, because that attack is currently set up in a way (minting MKR directly to the Flap contract) that perfectly avoids violating the `flapMkrConsistency` VFSM. I'd rather make that a separate PR as it isn't entirely trivial. After that, we can finally remove the unauthorized flap kick VFSM.

PR #168, opened 2020-04-06
Cleanups to model

ehildenb:
-   Bounds checks on `Vow` contract which enforce non-negative number, and cleanup of side-conditions in some cases.
-   Make sure we aren't overwriting the existing function-call events with custom events that are logged.
-   Add in "unimplemented" custom log assertions.
-   Make sure we have a post-state even when a violation is found (for making post-state delta for assertion generation).

### DIFF
--- a/kmcd-prelude.md
+++ b/kmcd-prelude.md
@@ -229,6 +229,9 @@ module KMCD-GEN
          <kmcd-state> STATE </kmcd-state>
          <kmcd-snapshots> ... (.List => ListItem(<kmcd-state> STATE </kmcd-state>)) </kmcd-snapshots>
 
+    rule <k> (. => snapshot) ~> #assert ... </k>
+         <kmcd-snapshots> ListItem(_) </kmcd-snapshots>
+
     syntax Int ::= #timeStepMax() [function]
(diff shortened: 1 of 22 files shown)


kmcd-prelude.md
```diff
@@ -229,6 +229,10 @@ module KMCD-GEN
          <kmcd-state> STATE </kmcd-state>
          <kmcd-snapshots> ... (.List => ListItem(<kmcd-state> STATE </kmcd-state>)) </kmcd-snapshots>
 
+    rule <k> (. => snapshot) ~> #assert ... </k>
+         <kmcd-snapshots> ListItem(_) </kmcd-snapshots>
+      [priority(60)]
```
kmbarry1, line 234: Might be good to add a comment explaining what the downstream purpose of lowering the priority here is, otherwise it's super-nonobvious to someone reading the code later on what it's doing.

ehildenb: I just ended up removing the `priority(60)` attribute altogether by adding a new state `#assert-failure` which `#assert` rewrites to if it doesn't get cleared from the K cell. This way we can avoid introducing the new feature into the semantics. See 24d8558.

PR #169, opened 2020-04-06
Handle more assertions, fixes to violation checking

ehildenb:
This should wait for #168 and #167 before merge (to address merge conflicts), but morally should be reviewed separately.

-   Simplifies the logic of flattening collections in several places in `mcd-pyk.py`.
-   Fixes the violation checking in `mcd-pyk.py`.
-   Handles more assertion generation when there is a mapping update during execution.

An example of a mapping update looks like:

```
    assertTrue( vat.can(address(bobby), address(end)) != 0 );
    assertTrue( vat.can(address(bobby), address(alice)) != 0 );
    assertTrue( vat.can(address(bobby), address(goldFlip)) != 0 );
    assertTrue( vat.can(address(bobby), address(pot)) != 0 );
    assertTrue( vat.can(address(alice), address(end)) != 0 );
    assertTrue( vat.can(address(alice), address(goldFlip)) != 0 );
    assertTrue( vat.can(address(alice), address(pot)) != 0 );
```

I wasn't sure how to do the double-access of `address(alice)` followed by `address(pot)` in the `can` map, so for now I just put both arguments together. This can easily be changed.

### DIFF
--- a/mcd-pyk.py
+++ b/mcd-pyk.py
@@ -48,23 +48,39 @@ def sanitizeBytes(kast):
         return _kast
     return pyk.traverseBottomUp(kast, _sanitizeBytes)
 
-def fromListItem(input):
-    if pyk.isKApply(input) and input['label'] == 'ListItem':
+def fromItem(input):
+    if pyk.isKApply(input) and input['label'] in [ 'ListItem' , 'SetItem' ]:
         return input['args'][0]
     return input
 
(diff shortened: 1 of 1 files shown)


gbalabasquer:
`vat.can(address(bobby), address(end))` is correct. Basically is checking if `bobby` has `hope`d `end`.

PR #171, opened 2020-04-08
Add flop blocking attack and property for detection

kmbarry1:
Not really an attack, just gets the system into a state where `flopping` should happen but can't. If you change the `heal` call to a `kiss` call, you'll see that the second `flop` gets successfully kicked off.

### DIFF
--- a/kmcd-props.md
+++ b/kmcd-props.md
@@ -23,9 +23,9 @@ Measurables
 ```k
     syntax Event ::= Measure
     syntax Measure ::= Measure () [function]
-                     | Measure ( debt: Rat , controlDai: Map , potChi: Rat , potPie: Rat , sumOfScaledArts: Rat , vice: Rat , endDebt: Rat , sumOfAllFlapLots: Rat , dai: Map , sumOfAllFlapBids: Rat , mkrBalances: Map ) [klabel(LogMeasure), symbol]
+                     | Measure ( debt: Rat , controlDai: Map , potChi: Rat , potPie: Rat , sumOfScaledArts: Rat , vice: Rat , endDebt: Rat , sumOfAllFlapLots: Rat , dai: Map , sumOfAllFlapBids: Rat , mkrBalances: Map, ash: Rat ) [klabel(LogMeasure), symbol]
  // ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
-    rule [[ Measure() => Measure(... debt: DEBT, controlDai: controlDais(keys_list(VAT_DAIS)), potChi: POT_CHI, potPie: POT_PIE, sumOfScaledArts: calcSumOfScaledArts(VAT_ILKS, VAT_URNS), vice: VAT_VICE, endDebt: END_DEBT, sumOfAllFlapLots: sumOfAllFlapLots(FLAP_BIDS), dai: VAT_DAIS, sumOfAllFlapBids: sumOfAllFlapBids(FLAP_BIDS), mkrBalances: mkrBalances()) ]]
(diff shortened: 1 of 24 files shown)


kmbarry1:
Added a property to catch "blocked flop" situations. Sorry for kind of implementing it in a weird way (using binary digits of an `Int` for an `Int` to `Bool` mapping), but I couldn't get it to work using a `Map` (parsing errors).

Open to guidance on how to do this less weirdly, but at least this works.

kmcd-props.md
```diff
@@ -23,9 +23,9 @@ Measurables
 ```k
     syntax Event ::= Measure
     syntax Measure ::= Measure () [function]
-                     | Measure ( debt: Rat , controlDai: Map , potChi: Rat , potPie: Rat , sumOfScaledArts: Rat , vice: Rat , endDebt: Rat , sumOfAllFlapLots: Rat , dai: Map , sumOfAllFlapBids: Rat , mkrBalances: Map ) [klabel(LogMeasure), symbol]
+                     | Measure ( debt: Rat , controlDai: Map , potChi: Rat , potPie: Rat , sumOfScaledArts: Rat , vice: Rat , endDebt: Rat , sumOfAllFlapLots: Rat , dai: Map , sumOfAllFlapBids: Rat , mkrBalances: Map, ash: Rat ) [klabel(LogMeasure), symbol]
  // ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
ehildenb, line 27:
```suggestion
 // -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

tests/attacks/blocked-flop.mcd
```diff
@@ -0,0 +1,36 @@
...
+// ---------------
+
+// Set sump (determines DAI bid size for flop auctions)
+transact ADMIN Vow . file sump 50
+
+// Set dump (determines initial MKR lot size for flop auctions)
+transact ADMIN Vow . file dump 30
+
+// Set tau (flop auction lifetime in seconds)
+transact ADMIN Flop . file tau 3600
```
ehildenb, line 14: Would these make sense in the DEPLOY-PRELUDE instead?

tests/attacks/blocked-flop.mcd
```diff
@@ -0,0 +1,36 @@
...
+
+// Use suck to give Bobby some DAI, which conveniently also creates enough free, unbacked debt for 2 flop auctions
+transact ADMIN Vat . suck Vow "Bobby" 100
+
+// Start a flop auction
+transact "Bobby" Vow . flop
+
+// Bobby must hope on the Flop contract to be able to bid
+transact "Bobby" Vat . hope Flop
+
```
ehildenb, line 24: I think this one makes sense in the ATTACK-PRELUDE.

kmbarry1: Hmmm seems that we've already put a bunch of `hope` calls in there anyway, so might as well stay consisten.

tests/attacks/blocked-flop.mcd
kmbarry1: I'm more torn on this one...on one hand, the system would never be "live" without these being initialized, on the other hand I like how having them in the attack file is both self-documenting and robust (i.e. even if someone adds a default setting in the DEPLOY-PRELUDE, no changes to this file are necessary).

To muddy the waters further, `hump` and `bump` are currently set in the `ATTACK-PRELUDE` but should arguably be set in the `DEPLOY-PRELUDE`.

I guess what I'll do is move parameter setting into the `DEPLOY-PRELUDE` and just put a comment in this attack file about what values it's assuming. Slightly more fragile but keeps lines of code down.

kmbarry1:
(comments addressed)

PR #173, opened 2020-04-14
Separate out representation of Wad/Rad/Ray

ehildenb:
This is to make it easier to change out the underlying representation of Wad/Rad/Ray in the semantics without having to change every call-site. It will also provide stronger information at each call-site about the sorts of all quantities.

### DIFF
--- a/cat.md
+++ b/cat.md
@@ -126,7 +126,7 @@ Cat Semantics
           ~> call Vow . fess TAB
           ~> call Flip ILK . kick URN VOWADDR (TAB *Rad Ray2Rad(CHOP)) LOT 0Rad
           ~> emitBite ILK URN LOT ART TAB)
-          (ART *Rate RATE))
+          (ART *RateWad RATE))
           (minWad(URNART, (LOT *Wad URNART) /Wad INK)))
           (minWad(INK, LUMP))
          ...
@@ -137,7 +137,7 @@ Cat Semantics
          <vat-ilks> ... ILK |-> Ilk(... rate: RATE, spot: SPOT) ... </vat-ilks>
          <vat-urns> ... { ILK, URN } |-> Urn(... ink: INK, art: URNART ) ... </vat-urns>
          <cat-vow> VOWADDR </cat-vow>
-      requires (INK *Rate SPOT) <Rad (URNART *Rate RATE)
+      requires (INK *RateWad SPOT) <Rad (URNART *RateWad RATE)
 
     syntax CatAuthStep ::= "cage" [klabel(#CatCage), symbol]
  // --------------------------------------------------------
--- a/cat.md
+++ b/cat.md
@@ -101,11 +101,11 @@ The parameters controlled by governance are:
 
     rule <k> Cat . file chop ILKID CHOP => . ... </k>
          <cat-ilks> ... ILKID |-> Ilk ( ... chop: (_ => CHOP) ) ... </cat-ilks>
-      requires CHOP >=Rat 0
+      requires CHOP >=Ray 0Ray
 
     rule <k> Cat . file lump ILKID LUMP => . ... </k>
          <cat-ilks> ... ILKID |-> Ilk ( ... lump: (_ => LUMP) ) ... </cat-ilks>
-      requires LUMP >=Rat 0
+      requires LUMP >=Wad 0Wad
 ```
 
 **NOTE**: `flip` is not fileable since we are assuming a unique liquidator for each ilk.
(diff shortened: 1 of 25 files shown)


kmcd-data.md
```diff
@@ -0,0 +1,237 @@
...
+ // ----------------------------------------------
+    rule minRay(W1, W2) => minRat(W1, W2)
+    rule maxRay(W1, W2) => maxRat(W1, W2)
+
+    syntax Rad ::= minRad ( Rad , Rad ) [function]
+                 | maxRad ( Rad , Rad ) [function]
+ // ----------------------------------------------
+    rule minRad(W1, W2) => minRat(W1, W2)
+    rule maxRad(W1, W2) => maxRat(W1, W2)
+
```
ehildenb, line 124:
```suggestion
```

cat.md
```diff
@@ -120,22 +122,22 @@ Cat Semantics
           => #fun(LOT
           => #fun(ART
           => #fun(TAB
-          => call Vat . grab ILK URN THIS VOWADDR (-1 *Rat LOT) (-1 *Rat ART)
+          => call Vat . grab ILK URN THIS VOWADDR (0Wad -Wad LOT) (0Wad -Wad ART)
           ~> call Vow . fess TAB
-          ~> call Flip ILK . kick URN VOWADDR (TAB *Rat CHOP) LOT 0
+          ~> call Flip ILK . kick URN VOWADDR (TAB *Rad Ray2Rad(CHOP)) LOT 0Rad
```
kmbarry1, line 127: I think `(TAB *Rad Ray2Rad(CHOP))` isn't quite what we want...this conceptually results in some sort of "RadRad" (90 digit fixed point).

I think what we really want is `rmul(TAB, CHOP)`. I guess we'll need to add an `rmul` for `Rad` left args:
```
syntax Rad ::= rmul ( Rad , Ray ) [function]
// etc
```
Hopefully that doesn't clash with the existing `rmul` defined for `Wads`. If there's an easy way to do parametric typing in `K`, the really "correct" way to define `rmul` would be something like (making up a syntax here):
```
template <numeric_sort T>
syntax T ::= rmul ( T , Ray ) [function]
// def
```
so `rmul` always returns the type of its first argument.

kmcd-data.md
```diff
@@ -0,0 +1,294 @@
...
+    rule W1   <Rad W2 => W1   <Rat W2
+    rule W1  >=Rad W2 => W1  >=Rat W2
+    rule W1   >Rad W2 => W1   >Rat W2
+    rule W1  ==Rad W2 => W1  ==Rat W2
+    rule W1 =/=Rad W2 => W1 =/=Rat W2
+```
+
+```k
+    syntax Rad ::= Wad "*RateWad" Ray [function]
+                 | Rad "*RateRad" Ray [function]
```
kmbarry1, line 180: `*RateRad` doesn't seem to be used anywhere, and it really shouldn't be necessary...in that case, we can get rid of it and `*RateWad` can just be `*Rate`.

kmbarry1: Although IIRC, you said there was some reason for this on the call...but I've managed to forget it between this morning and tonight.

But `*RateRad` wouldn't result in a `Rad` anyway...it would result in a (45+27) digit fixed point number. What you really want, I think, is the `rmul(Rad, Ray)` I mentioned earlier. If `K` really won't do parametric typing in a nice way (and since you aren't already using it, I suspect it doesn't), we can have to operators:
`syntax Wad ::= wrmul(Wad, Ray)` (or call it `*WadRay`)
`syntax Rad ::= rrmul(Rad, Ray)` (or call it `*RadRay`)

kmcd-prelude.md
```diff
@@ -102,9 +102,9 @@ module KMCD-PRELUDE
 
          // Initialize Spot for gold
          transact ADMIN Spot . init     "gold"
-         transact ADMIN Spot . setPrice "gold" 3 ether
-         transact ADMIN Spot . file       mat "gold" 1
-         transact ADMIN Spot . file       par 1
+         transact ADMIN Spot . setPrice "gold" Int2Wad(3 ether)
```
kmbarry1, line 105: This is going to do a weird thing when we switch to `Int`. I guess it's okay, we can fix it then (with the `Int2Wad` invocation, it's more in line with the intent to just change `3 ether` to `3`).

kmbarry1: Would be nice to be able to e.g. write `3 * 1Wad`.

kmcd-data.md
kmbarry1: Ah it was hidden in the kmcd-props.md diff (not loaded by default)

kmbarry1: Looking at how it's being used, some sort of `rmul` is definitely the right way to express this.

cat.md
ehildenb: I've switched this to `rmulRad`.

kmcd-data.md
ehildenb: Same here, I've switch to `rmulRad`. I can't just have it be `rmul` until we actually switch over to separate representations for each type (next PR), because the sort inferencer will get confused.

kmcd-prelude.md
ehildenb: Later it will just be `wad(3 ether)`. But for now I'm using the `Int2Wad` converter to keep the sort inferencer happy.

ehildenb: I've gone ahead and included the change that puts `wad(...)`, `ray(...)`, `rad(...)` instead of the `Int2*` functions.

PR #182, opened 2020-05-05
Clean separation of representation of Wad/Ray/Rad

ehildenb:
There were still some issues with changing the underlying representation of these data-types, so this refactors to make it easier. Basic changes include:

-   Removal of `0{Wad,Ray,Rad}` and `1{Wad,Ray,Rad}` in favor of `{wad,ray,rad}(0)` and `{wad,ray,rad}(1)`.
-   Avoid names `WAD`, `RAY`, and `RAD` for variable names.
-   Switch away from directly putting rationals in configurations of `flip`, `flap`, and `flop`.
-   Add quantities `WAD`, `RAY`, and `RAD` for convenience later.

### DIFF
--- a/kmcd-data.md
+++ b/kmcd-data.md
@@ -23,6 +23,9 @@ We model everything with arbitrary precision rationals, but use sort information
 -   `Rad`: result of multiplying `Wad` and `Ray` (highest precision). Represented in implementation as 1e45 fixed point.
 
 ```k
+    syntax Value ::= Wad | Ray | Rad
+ // --------------------------------
+
     syntax Int ::= "WAD" | "RAY" | "RAD"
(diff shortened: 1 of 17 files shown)


kmcd-data.md
```diff
@@ -55,34 +64,26 @@ We model everything with arbitrary precision rationals, but use sort information
```
kmbarry1, line 64: Just above this line (commenting here since I can't comment outside of the diff), there are defs for `0Wad`, `1Wad`,  etc...seems like the rest of this PR systematically removes those in favor of `wad(0)`, etc; is there a reason to keep them?

ehildenb: Good catch, I've removed them.

PR #184, opened 2020-05-06
Fixed point representation for Wad/Ray/Rad

ehildenb:
This switches the underlying type of Wad/Ray/Rad to `FInt`, a new fixed-point integer representation.


fixed-int.md
```diff
@@ -0,0 +1,44 @@
+```k
+module FIXED-INT
+    imports INT
+    imports BOOL
+
+    syntax FInt ::= "(" FInt ")"                   [bracket]
+                  | FInt ( value: Int , one: Int ) [klabel(FInt), symbol]
+ // ---------------------------------------------------------------------
+    // rule #Ceil(FInt(... value: V, one: O)) => { V >=Int 0 andBool O >Int 0 #Equals true } [anywhere, symbolic]
```
kmbarry1, line 9: So this is commented out...will it eventually be uncommented, or can we get rid of it?

If we're keeping it:
1) is the "`#Equals true`" part really necessary? Looks like `V >=Int 0 andBool O >Int 0` should already evaluate to a boolean.
2) as I was once told on a topology homework assignment, the letter O is an unfortunate choice for a variable name...

fixed-int.md
```diff
@@ -0,0 +1,44 @@
...
+    // Operations always produce the width of the left integer.
+    syntax FInt ::= FInt "*FInt" FInt [function]
+                  | FInt "/FInt" FInt [function]
+                  | FInt "^FInt"  Int [function]
+                  > FInt "+FInt" FInt [function]
+                  | FInt "-FInt" FInt [function]
+ // --------------------------------------------
+    rule FInt(V1, O1) *FInt FInt(V2, O2) => FInt((V1 *Int V2) /Int O2, O1)
+    rule FInt(V1, O1) /FInt FInt(V2, O2) => FInt((V1 *Int O2) /Int V2, O1) requires V2 =/=Int 0
+    rule FInt(V1, O1) ^FInt E            => FInt(V1 ^Int E, O1)
```
kmbarry1, line 20: This doesn't seem right, e.g.:

`(1 in width 100)^2 = FInt(100, 100) ^FInt 2 = FInt(100 ^Int 2, 100) = FInt(10000, 100) = (100 in width 100)`

but I'd expect the result to just equal `(1 in width 100)`.

A correct option could be:

`rule FInt(V1, O1) ^FInt E            => FInt((V1 ^Int E) /Int (O1 ^Int (E -Int 1)), O1) requires E >Int 0`
`rule FInt(V1, O1) ^Fint E            => FInt(O1, O1) requires E ==Int 0`

kmcd-data.md
```diff
@@ -137,38 +166,40 @@ We model everything with arbitrary precision rationals, but use sort information
...
+    rule FInt(R1, RAD) /Rate FInt(R2, RAY) => FInt(R1 /Int R2, WAD) [owise]
 
-    syntax Rad ::= rmul ( Rad , Ray ) [function]
- // --------------------------------------------
-    rule rmul(rad(R1), ray(R2)) => rad(R1 *Rat R2)
+    // syntax Wad ::= rmul ( Wad , Ray ) [function]
+    // syntax Rad ::= rmul ( Rad , Ray ) [function]
+    syntax FInt ::= rmul ( FInt , FInt ) [function]
+ // -----------------------------------------------
+    rule rmul(FInt(W, WAD), FInt(R, RAY)) => FInt((W *Int R) /Int RAY, WAD)
```
kmbarry1, line 191: I think these two lines could be replaced with:
`rule rmul(FInt(V1, O1), FInt(V2, O2)) => FInt(V1, O1) *FInt FInt(V2,O2)`
given how `*FInt` was defined.

Although maybe it's better to be explict?

kmcd-data.md
```diff
@@ -137,38 +166,40 @@ We model everything with arbitrary precision rationals, but use sort information
...
-    syntax Wad ::= rmul ( Wad , Ray ) [function]
- // --------------------------------------------
-    rule rmul(wad(R1), ray(R2)) => wad(R1 *Rat R2)
+    rule FInt(_ , _  ) /Rate ray(0)        => wad(0)
+    rule FInt(R1, RAD) /Rate FInt(R2, RAY) => FInt(R1 /Int R2, WAD) [owise]
 
-    syntax Rad ::= rmul ( Rad , Ray ) [function]
- // --------------------------------------------
-    rule rmul(rad(R1), ray(R2)) => rad(R1 *Rat R2)
+    // syntax Wad ::= rmul ( Wad , Ray ) [function]
```
kmbarry1, line 187: Should these commented lines be removed?

kmcd-data.md
```diff
@@ -137,38 +166,40 @@ We model everything with arbitrary precision rationals, but use sort information
...
+    // syntax Rad ::= rmul ( Rad , Ray ) [function]
+    syntax FInt ::= rmul ( FInt , FInt ) [function]
+ // -----------------------------------------------
+    rule rmul(FInt(W, WAD), FInt(R, RAY)) => FInt((W *Int R) /Int RAY, WAD)
+    rule rmul(FInt(W, RAD), FInt(R, RAY)) => FInt((W *Int R) /Int RAY, RAD)
 
     syntax Ray ::= rdiv ( Ray , Rad ) [function]
```
kmbarry1, line 194: In the codebase, `rdiv` is a thing that takes a `Ray` for its second arg and returns the type of the first arg.

kmbarry1: Not sure if this is being used anywhere...

kmcd-data.md
```diff
@@ -137,38 +166,40 @@ We model everything with arbitrary precision rationals, but use sort information
...
  // --------------------------------------------
-    rule rdiv(ray(R1), rad(R2)) => ray(R1 /Rat R2)
+    rule rdiv(FInt(_ , RAY), rad(0))        => ray(0)
+    rule rdiv(FInt(R1, RAY), FInt(R2, RAD)) => FInt((R1 *Int RAY) /Int R2, RAY) [owise]
 
     syntax Ray ::= wdiv ( Ray , Wad ) [function]
  // --------------------------------------------
-    rule wdiv(ray(R1), wad(R2)) => ray(R1 /Rat R2)
+    rule wdiv(FInt(_ , RAY), wad(0))        => ray(0)
+    rule wdiv(FInt(R1, RAY), FInt(R2, WAD)) => FInt((R1 *Int WAD) /Int R2, RAY) [owise]
```
kmbarry1, line 202: Could use `/FInt` here too if desired.

fixed-int.md
ehildenb: Well, we _should_ have this definedness axiom present, but right now the LLVM backend is rejecting it. We have to wait for some changes to the way functions work in K though before it can stop rejecting it.

So I'll remove the axiom, and I've renamed all the variables `O* => ONE*` and `V* => VALUE*`.

fixed-int.md
ehildenb: Alright, I've corrected this and added a case for `E <Int 0` as well.

kmcd-data.md
ehildenb: I just made `rmul`, `rdiv`, and `wdiv` directly go to the underlying `FInt` operators, except in division by zero case.

This means we don't get strong type-checking at uses of these operators (which we didn't get anyway), but we get the correct implementations (where the width of the first operand is always what's returned).

kmcd-data.md
ehildenb: Done.

kmcd-data.md
ehildenb: Done.

fixed-int.md
ehildenb: Would appreciate you double-checking my work here though, because it can be tricky.

ehildenb:
This is now blocked on a K frontend issue: https://github.com/kframework/k/issues/1273

fixed-int.md
```diff
@@ -0,0 +1,87 @@
...
+                  | FInt "/FInt" FInt [function]
+                  | FInt "^FInt"  Int [function]
+                  > FInt "+FInt" FInt [function]
+                  | FInt "-FInt" FInt [function]
+ // --------------------------------------------
+    rule FInt(VALUE1, ONE1) *FInt FInt(VALUE2, ONE2) => FInt((VALUE1 *Int VALUE2) /Int ONE2, ONE1)
+    rule FInt(VALUE1, ONE1) /FInt FInt(VALUE2, ONE2) => FInt((VALUE1 *Int ONE2) /Int VALUE2, ONE1) requires VALUE2 =/=Int 0
+    rule FInt(VALUE, ONE) ^FInt E                    => FInt((VALUE ^Int E) /Int (ONE ^Int (E -Int 1)), ONE)          requires E  >Int 0
+    rule FInt(VALUE, ONE) ^FInt E                    => 1FInt(ONE)                                                    requires E ==Int 0
+    rule FInt(VALUE, ONE) ^FInt E                    => FInt((ONE ^Int (1 -Int E)) /Int (VALUE ^Int (0 -Int E)), ONE) requires E  <Int 0
```
kmbarry1, line 57: This seems correct, after a few minutes inspection, at least.

PR #188, opened 2020-05-25
Run MKR testing harness on CI

ehildenb:
This just sets up the MKR testing harness for https://github.com/makerdao/mkr-mcd-spec-sol-tests to run on CI (just to make sure we have all the dependencies we need in place).

Next PR will update to actually generate the tests as specified and run them.


ehildenb:
Messed up the submodules.

PR #190, opened 2020-05-29
Quick fix for Makefile

ehildenb:


ehildenb:
The reason is that since the transfer of the repo to the `makerdao/` organization, those lines do not work anymore (because it's not the RV repo). So I wanted to disable them because I keep getting notifications about the failure to deploy to GitHub pages, but I want to leave them as documentation about how to deploy to GitHub pages on your guys repo.

PR #197, opened 2020-07-01
Updates before adding constructors/deploy logic

ehildenb:
-   Rename all occurances of `ILKID` to `ILK_ID` for consistency (easier grep-ability).
-   Add file parameters to some of the contracts that were not present before.
-   Some contracts get "default initialization" of their data-structures. Ideally we would have a solution to always do this, but we don't. So for now, I've handled the few cases that show up.
-   Removes all the `Unused variable` warnings that show up with newest K.

### DIFF
--- a/jug.md
+++ b/jug.md
@@ -103,7 +103,6 @@ Jug Semantics
          <jug-ilks> ... ILK_ID |-> Ilk ( ... duty: ray(0) => ray(1), rho: _ => NOW ) ... </jug-ilks>
 
     rule <k> Jug . init ILK_ID ... </k>
-         <current-time> NOW </current-time>
          <jug-ilks> JUG_ILKS => JUG_ILKS [ ILK_ID <- Ilk ( ... duty: ray(0) , rho: 0 ) ] </jug-ilks>
       requires notBool ILK_ID in_keys(JUG_ILKS)
 ```
--- a/jug.md
+++ b/jug.md
@@ -105,6 +105,7 @@ Jug Semantics
     rule <k> Jug . init ILK ... </k>
          <current-time> NOW </current-time>
          <jug-ilks> JUG_ILKS => JUG_ILKS [ ILK <- Ilk ( ... duty: ray(0) , rho: 0 ) ] </jug-ilks>
+      requires notBool ILK in_keys(JUG_ILKS)
 ```
 
 ```k
--- a/jug.md
+++ b/jug.md
@@ -101,6 +101,10 @@ Jug Semantics
     rule <k> Jug . init ILK => . ... </k>
          <current-time> NOW </current-time>
          <jug-ilks> ... ILK |-> Ilk ( ... duty: ray(0) => ray(1), rho: _ => NOW ) ... </jug-ilks>
+
+    rule <k> Jug . init ILK ... </k>
+         <current-time> NOW </current-time>
+         <jug-ilks> JUG_ILKS => JUG_ILKS [ ILK <- Ilk ( ... duty: ray(0) , rho: 0 ) ] </jug-ilks>
 ```
 
 ```k
--- a/jug.md
+++ b/jug.md
@@ -76,8 +76,8 @@ These parameters are controlled by governance:
                      | "base" Ray
                      | "vow-file" Address
  // -------------------------------------
-    rule <k> Jug . file duty ILKID DUTY => . ... </k>
-         <jug-ilks> ... ILK |-> Ilk ( ... duty: (_ => DUTY) , rho: RHO ) ... </jug-ilks>
+    rule <k> Jug . file duty ILK_ID DUTY => . ... </k>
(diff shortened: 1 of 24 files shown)


jug.md
```diff
@@ -98,18 +98,22 @@ Jug Semantics
...
     syntax JugAuthStep ::= "init" String
  // ------------------------------------
-    rule <k> Jug . init ILK => . ... </k>
+    rule <k> Jug . init ILK_ID => . ... </k>
          <current-time> NOW </current-time>
-         <jug-ilks> ... ILK |-> Ilk ( ... duty: ray(0) => ray(1), rho: _ => NOW ) ... </jug-ilks>
+         <jug-ilks> ... ILK_ID |-> Ilk ( ... duty: ray(0) => ray(1), rho: _ => NOW ) ... </jug-ilks>
+
+    rule <k> Jug . init ILK_ID ... </k>
+         <jug-ilks> JUG_ILKS => JUG_ILKS [ ILK_ID <- Ilk ( ... duty: ray(0) , rho: 0 ) ] </jug-ilks>
```
kmbarry1, line 106: I'm a little confused by the syntax here (not saying it's wrong, just wondering why things are the way they are). Naively, I'd expect something like:
```
<jug-ilks> JUG_ILKS => JUG_ILKS ( ILK_ID |-> Ilk (... duty: ray(0) , rho: 0 ) ) </jug-ilks>
```
How is the `[ K <- V ]` syntax different from the `K |-> V` syntax?

kmcd-props.md
```diff
@@ -126,13 +126,13 @@ Art of an ilk = Sum of all urn art across all users for that ilk.
 ```k
     syntax Wad ::= sumOfUrnArt(Map, String, Wad) [function, functional]
  // -------------------------------------------------------------------
-    rule sumOfUrnArt( {ILKID , ADDR} |-> Urn (... art: ART) URNS, ILKID', SUM)
-      => #if ILKID ==K ILKID'
-            #then sumOfUrnArt( URNS, ILKID', SUM +Wad ART)
-            #else sumOfUrnArt( URNS, ILKID', SUM)
+    rule sumOfUrnArt( {ILK_ID , _ADDR} |-> Urn (... art: ART) URNS, ILK_ID', SUM)
```
kmbarry1, line 129: Could we just use `_` instead of `_ADDR`?

kmcd-props.md
```diff
@@ -391,9 +391,9 @@ The property checks if a successful `Pot . join` is preceded by a `TimeStep` mor
 
     rule derive(flopBlockCheck(... embers: EMBERS, dented: DENTED), FlopKick(_, _, BID, _)) => flopBlockCheck(... embers: EMBERS +Rad BID, dented: DENTED)
 
-    rule derive(flopBlockCheck(... embers: EMBERS, dented: DENTED), LogNote(_ , Flop . dent ID _ BID)) => flopBlockCheck(... embers: EMBERS -Rad BID, dented: DENTED +Int (2 ^Int ID))
+    rule derive(flopBlockCheck(... embers: EMBERS, dented: DENTED), LogNote(_ , Flop . dent  ID _  BID)) => flopBlockCheck(... embers: EMBERS -Rad BID, dented: DENTED +Int (2 ^Int ID))
         requires ( ( ( DENTED /Int ( 2 ^Int ID ) ) modInt 2 ) ==Int 0 )
-    rule derive(flopBlockCheck(... embers: EMBERS, dented: DENTED), LogNote(_ , Flop . dent ID _ BID)) => flopBlockCheck(... embers: EMBERS, dented: DENTED) [owise]
+    rule derive(flopBlockCheck(... embers: EMBERS, dented: DENTED), LogNote(_ , Flop . dent _ID _ _BID)) => flopBlockCheck(... embers: EMBERS, dented: DENTED) [owise]
```
kmbarry1, line 396: Same sort of thing here...if we're using `_` for `LOT` already, why not just `_ _ _` instead of `_ID` and `_BID`? Just because they're named in the rule above?

jug.md
ehildenb: The operator `_Map_` is map union, and `K |-> V` is a constructor of sort `Map`, so when you do `MAP => MAP (K |-> V)` (which looks like `MAP => _Map_(MAP, _|->_(K, V))` in AST form), you're asserting that the maps `MAP` and `(K |-> V)` have disjoint keys. _If_ that assumption is broken, then it's undefined (entire thing is #False in Haskell backend, and runtime error in LLVM backend).

The operator `_[_<-_]` is map update, so if you do `MAP => MAP [ K <- V ]` and `MAP` already contains key `K`, it will just override that entry in `MAP` with `K |-> V`, instead of crashing.

kmcd-props.md
ehildenb: Done.

kmcd-props.md
ehildenb: Done

PR #198, opened 2020-07-01
Semantics of deployment process

ehildenb:
Instead of "manually" hardcoding the deployment process in the DEPLOY-PRELUDE, here we add constructors and call them in the same way that DssDeploy is called.

-   Add `constructor` for all contracts, mimicking what the contracts constructor does.
-   Add `file` for contracts missing it.
-   Add deployment logic to `kmcd.md`, use this deployment logic for `DEPLOY-PRELUDE`.
-   Add tests `tests/fixes/{deploy,attack}Prelude.mcd` which just test the post-state after running these preludes.


kmcd-prelude.md
```diff
@@ -16,29 +16,18 @@ module KMCD-PRELUDE
...
+         // ----------------
 
-         // Authorize End for Cat/Pot
-         transact ADMIN Cat . rely End
+         transact ADMIN Gem "MKR" . init
+         transact ADMIN Deploy . deploy Gem "MKR"
 
-         // Authorize Vow for Flap
-         transact ADMIN Flap . rely Vow
+         transact ADMIN GemJoin "MKR" . constructor Vat Gem "MKR"
```
kmbarry1, line 25: We shouldn't need a `GemJoin` for MKR...it is not used as collateral (and never should be due to its special role in the protocol).

kmbarry1: Heh, seems like I missed this in prior reviews :)

kmcd-prelude.md
```diff
@@ -16,29 +16,18 @@ module KMCD-PRELUDE
...
+         transact ADMIN Gem     "MKR" . initUser GemJoin "MKR"
+         transact ADMIN Deploy . deployCollateral "MKR" GemJoin "MKR"
 
-         // Authorize Vow for Flop
-         transact ADMIN Flop . rely Vow
+         transact ADMIN Gem "MKR" . initUser Vow
```
kmbarry1, line 29: I don't think the `Vow` needs an MKR balance? (line below is needed for the `Flap` for sure, though)

kmcd-prelude.md
```diff
@@ -92,7 +67,12 @@ module KMCD-PRELUDE
 
          // "gold" collateral and joiner
          transact ADMIN Gem "gold" . init
-         transact ADMIN GemJoin "gold" . init
+         transact ADMIN GemJoin "gold" . constructor Vat Gem "gold"
+         transact ADMIN Gem     "gold" . initUser GemJoin "gold"
+         transact ADMIN Deploy . deployCollateral "gold" GemJoin "gold"
+
+         transact ADMIN Gem "gold" . initUser Vow
```
kmbarry1, line 74: I don't think either of these two lines are necessary? Only the `GemJoin` of "gold" needs a `Gem` balance, unless transferring "gold" to either the `Vow` or `Flap` is part of an attack (which it is not the case as far as I can recall)?

kmcd.md
```diff
@@ -72,6 +72,115 @@ State Storage/Revert Semantics
...
+          ~> Pot  . rely End
+          ~> Spot . rely End
+         ...
+         </k>
+
+    syntax DeployStep ::= "deployCollateral" String Address
+ // -------------------------------------------------------
+    rule <k> Deploy . deployCollateral ILK_ID JOIN
+          => call Flip ILK_ID . constructor Vat
+          // ~> call Spot . file pip ILK_ID PIP
```
kmbarry1, line 173: Is this commented out because we don't model the `pips` as contracts?

ehildenb: Yes, that's correct.

kmcd-prelude.md
ehildenb: Indeed we don't need this, it's removed.

kmcd-prelude.md
ehildenb: Looks like for `lucash-flap-end` attacks, it's needed (because we do some hacky stuff to setup the attack state).

So I've localized this call to those attack files and removed it from this prelude.

kmcd-prelude.md
ehildenb: Looks like they aren't. They're removed.

PR #201, opened 2020-07-08
Remove call from `Vow` in prelude

ehildenb:
This removes the last remaining invalid call in the prelude, and fixes that call in the `Vow` constructor.

Now, all calls in the preludes are from `ADMIN`, `ANYONE`, `"Bobby"`, or `"Alice"`, and not form internal MCD contracts (which doesn't make much sense because those contracts can't initiate calls).


tests/attacks/lucash-flap-end.mcd
```diff
@@ -2,6 +2,7 @@ STEPS ( DEPLOY-PRELUDE )
 STEPS ( ATTACK-PRELUDE )
 
 transact ADMIN Gem "MKR" . initUser Vow
+transact Vow Vat . hope Flap
```
kmbarry1, line 5: This should really be in the Vow's constructor, is there a reason it can't go there?

ehildenb: Ah, actually it is there! I missed that.

PR #205, opened 2020-07-14
Add error message for assertTrue in solidity tests

gbalabasquer:


ehildenb:
@gbalabasquer you'll also need to change `mcd-pyk.py`. Specifically, function `makeSolidityCall` and `buildAsserts`.

I would just keep a global counter in the python file, and have each assert print out something like `"assert NNN"` where `N` is the global counter number of each time it makes an assert. This will make it easier to see which assert failed.

I can do it instead if you like.

gbalabasquer:
@ehildenb ahh ok, if you can easily do it please go ahead, it was a long time since I've used python so you will probably do it much faster

PR #241, opened 2021-06-08
Simplifications for symbolic reasoning

ehildenb:
These are a few simplifications to the semantics I found which help my local work with symbolic reasoning.

- Make `wish` a non-contextual function, so that it's possible to use it over the pre/post states in proofs.
- Mark `wards` as functional.
- Add `domains.md` explicit import.
- `makeCall` no longer defaults to throwing an exception, because `checkLock` and `checkAuth` handle that.

### DIFF
--- a/fixed-int.md
+++ b/fixed-int.md
@@ -5,6 +5,8 @@ This module implements fixed point integers in K.
 Once it stabalizes, it should be upstreamed into the K Prelude.
 
 ```k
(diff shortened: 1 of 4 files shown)


0xverif (approved):
nice code reorg and improved wish function

0xverif (approved):
nice code reorg and improved wish function

PR #250, opened 2021-06-30
Liq 2.0

0xverif:
New K implementations of Abaci, Clip and Dog

### DIFF
--- a/abaci.md
+++ b/abaci.md
@@ -41,7 +41,7 @@ Abacus Configuration
     rule <k> Abacus ABACUS_ADDRESS . constructor ABACUS_TYPE => . ... </k>
          <msg-sender> MSGSENDER </msg-sender>
          ( .Bag
-        => <abacus>
+         => <abacus>
             <abacus-address> ABACUS_ADDRESS     </abacus-address>
             <abacus-type>    ABACUS_TYPE        </abacus-type>
             <abacus-wards>   SetItem(MSGSENDER) </abacus-wards>
@@ -82,10 +82,9 @@ File-able Fields
 
 These parameters are controlled by governance:
 
--   `tau`:  Seconds after auction start when the price reaches zeroe
+-   `tau`:  Seconds after auction start when the price reaches zero
 -   `step`: Length of time between price drops
 -   `cut`:  Per-step multiplicative factor
-
 
 ```k
     syntax AbacusAuthStep ::= "file" AbacusFile
@@ -115,6 +114,7 @@ These parameters are controlled by governance:
             <abacus-cut>     _ => CUT       </abacus-cut>
             ...
          </abacus>
+      requires CUT <=Ray ray(1)
 ```
 
 Abacus Semantics
@@ -126,47 +126,47 @@ Abacus Semantics
    syntax AbacusStep ::= "price" Ray Int
  // ------------------------------------
    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
-      <abacus>
-         <abacus-address> ABACUS_ADDRESS   </abacus-address>
-         <abacus-type>    "LinearDecrease" </abacus-type>
-         <abacus-tau>     ABACUS_TAU       </abacus-tau>
-         ...
-      </abacus>
-      requires DUR <=Int ABACUS_TAU
+         <abacus>
+            <abacus-address> ABACUS_ADDRESS   </abacus-address>
+            <abacus-type>    "LinearDecrease" </abacus-type>
+            <abacus-tau>     ABACUS_TAU       </abacus-tau>
+            ...
+         </abacus>
+      requires DUR <Int ABACUS_TAU
 
-    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ray(0) ... </k>
-      <abacus>
-         <abacus-address> ABACUS_ADDRESS   </abacus-address>
-         <abacus-type>    "LinearDecrease" </abacus-type>
-         <abacus-tau>     ABACUS_TAU       </abacus-tau>
-         ...
-      </abacus>
-        requires DUR >Int ABACUS_TAU
+    rule <k> Abacus ABACUS_ADDRESS . price _TOP DUR => ray(0) ... </k>
+         <abacus>
+            <abacus-address> ABACUS_ADDRESS   </abacus-address>
+            <abacus-type>    "LinearDecrease" </abacus-type>
+            <abacus-tau>     ABACUS_TAU       </abacus-tau>
+            ...
+         </abacus>
+      requires DUR >=Int ABACUS_TAU
 ```
 
 - `StairstepExponentialDecrease` abacus:
 
 ```k
    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ( ABACUS_CUT ^Ray ( DUR /Int ABACUS_STEP ) ) ) ... </k>
-      <abacus>
-         <abacus-address> ABACUS_ADDRESS                 </abacus-address>
-         <abacus-type>    "StairstepExponentialDecrease" </abacus-type>
-         <abacus-step>    ABACUS_STEP                    </abacus-step>
-         <abacus-cut>     ABACUS_CUT                     </abacus-cut>
-         ...
-      </abacus>
+         <abacus>
+            <abacus-address> ABACUS_ADDRESS                 </abacus-address>
+            <abacus-type>    "StairstepExponentialDecrease" </abacus-type>
+            <abacus-step>    ABACUS_STEP                    </abacus-step>
+            <abacus-cut>     ABACUS_CUT                     </abacus-cut>
+            ...
+         </abacus>
 ```
 
 - `ExponentialDecrease` abacus:
 
 ```k
    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => TOP *Ray ( ABACUS_CUT ^Ray DUR ) ... </k>
-      <abacus>
-         <abacus-address> ABACUS_ADDRESS        </abacus-address>
-         <abacus-type>    "ExponentialDecrease" </abacus-type>
-         <abacus-cut>     ABACUS_CUT            </abacus-cut>
-         ...
-      </abacus>
+         <abacus>
+            <abacus-address> ABACUS_ADDRESS        </abacus-address>
+            <abacus-type>    "ExponentialDecrease" </abacus-type>
+            <abacus-cut>     ABACUS_CUT            </abacus-cut>
+            ...
+         </abacus>
 ```
 
 ```k
--- a/abaci.md
+++ b/abaci.md
@@ -125,7 +125,7 @@ Abacus Semantics
 ```k
     syntax AbacusStep ::= "price" Ray Int
  // ------------------------------------
-    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP  *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
+    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
          <abacus>
             <abacus-address> ABACUS_ADDRESS   </abacus-address>
             <abacus-type>    "LinearDecrease" </abacus-type>
--- a/abaci.md
+++ b/abaci.md
@@ -61,7 +61,7 @@ Abacus Authorization
     rule [[ wards(Abacus ABACUS_ADDRESS) => WARDS ]] <abaci> <abacus-address> ABACUS_ADDRESS </abacus-address> <abacus-wards> WARDS </abacus-wards> ... </abaci>
 
     syntax AbacusAuthStep ::= WardStep
- // --------------------------------
+ // ----------------------------------
     rule <k> Abacus ABACUS_ADDRESS . rely ADDR => . ... </k>
          <abacus>
            <abacus-address>   ABACUS_ADDRESS              </abacus-address>
@@ -76,7 +76,6 @@ Abacus Authorization
            ...
          </abacus>
 ```
-
 
 File-able Fields
 ----------------
@@ -95,7 +94,7 @@ These parameters are controlled by governance:
     syntax AbacusFile ::= "tau"  Int
                         | "step" Int
                         | "cut"  Ray
- // --------------------------------------------
+ // --------------------------------
     rule <k> Abacus ABACUS_ADDRESS . file tau TAU => . ... </k>
          <abacus>
             <abacus-address> ABACUS_ADDRESS </abacus-address>
@@ -118,13 +117,14 @@ These parameters are controlled by governance:
          </abacus>
 ```
 
-
 Abacus Semantics
 ----------------
 
+- `LinearDecrease` abacus:
+
 ```k
    syntax AbacusStep ::= "price" Ray Int
- // -------------------------------------
+ // ------------------------------------
    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
       <abacus>
          <abacus-address> ABACUS_ADDRESS   </abacus-address>
@@ -132,8 +132,21 @@ Abacus Semantics
          <abacus-tau>     ABACUS_TAU       </abacus-tau>
          ...
       </abacus>
-      requires DUR >=Int ABACUS_TAU
+      requires DUR <=Int ABACUS_TAU
 
+    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ray(0) ... </k>
+      <abacus>
+         <abacus-address> ABACUS_ADDRESS   </abacus-address>
+         <abacus-type>    "LinearDecrease" </abacus-type>
+         <abacus-tau>     ABACUS_TAU       </abacus-tau>
+         ...
+      </abacus>
+        requires DUR >Int ABACUS_TAU
+```
+
+- `StairstepExponentialDecrease` abacus:
+
+```k
    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ( ABACUS_CUT ^Ray ( DUR /Int ABACUS_STEP ) ) ) ... </k>
       <abacus>
          <abacus-address> ABACUS_ADDRESS                 </abacus-address>
@@ -142,7 +155,11 @@ Abacus Semantics
          <abacus-cut>     ABACUS_CUT                     </abacus-cut>
          ...
       </abacus>
+```
 
+- `ExponentialDecrease` abacus:
+
+```k
    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => TOP *Ray ( ABACUS_CUT ^Ray DUR ) ... </k>
       <abacus>
          <abacus-address> ABACUS_ADDRESS        </abacus-address>
@@ -150,7 +167,6 @@ Abacus Semantics
          <abacus-cut>     ABACUS_CUT            </abacus-cut>
          ...
       </abacus>
-
 ```
 
 ```k
--- a/abaci.md
+++ b/abaci.md
@@ -123,9 +123,9 @@ Abacus Semantics
 - `LinearDecrease` abacus:
 
 ```k
-   syntax AbacusStep ::= "price" Ray Int
+    syntax AbacusStep ::= "price" Ray Int
  // ------------------------------------
-   rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
+    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP  *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
          <abacus>
             <abacus-address> ABACUS_ADDRESS   </abacus-address>
             <abacus-type>    "LinearDecrease" </abacus-type>
@@ -147,7 +147,7 @@ Abacus Semantics
 - `StairstepExponentialDecrease` abacus:
 
 ```k
-   rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ( ABACUS_CUT ^Ray ( DUR /Int ABACUS_STEP ) ) ) ... </k>
+    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ( ABACUS_CUT ^Ray ( DUR /Int ABACUS_STEP ) ) ) ... </k>
          <abacus>
             <abacus-address> ABACUS_ADDRESS                 </abacus-address>
             <abacus-type>    "StairstepExponentialDecrease" </abacus-type>
@@ -160,7 +160,7 @@ Abacus Semantics
 - `ExponentialDecrease` abacus:
 
 ```k
-   rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => TOP *Ray ( ABACUS_CUT ^Ray DUR ) ... </k>
+    rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => TOP *Ray ( ABACUS_CUT ^Ray DUR ) ... </k>
          <abacus>
             <abacus-address> ABACUS_ADDRESS        </abacus-address>
             <abacus-type>    "ExponentialDecrease" </abacus-type>
--- a/abaci.md
+++ b/abaci.md
@@ -85,7 +85,7 @@ These parameters are controlled by governance:
 
 -   `tau`:  Seconds after auction start when the price reaches zeroe
 -   `step`: Length of time between price drops
--   `cut`:  Per-step multiplicative facto
+-   `cut`:  Per-step multiplicative factor
 
 
 ```k
--- a/abaci.md
+++ b/abaci.md
@@ -1,0 +1,158 @@
+```k
+requires "kmcd-driver.md"
+
+module ABACI
+    imports KMCD-DRIVER
+```
+
+Abacus Configuration
+-----------------
+
+```k
+    configuration
+    <abaci>
+      <abacus multiplicity="*" type="Map">
+         <abacus-address> 0:Address </abacus-address>
+         <abacus-type>    ""        </abacus-type>
+         <abacus-wards>   .Set      </abacus-wards>
+         <abacus-tau>     0         </abacus-tau>
+         <abacus-step>    0         </abacus-step>
+         <abacus-cut>     ray(0)    </abacus-cut>
+      </abacus>
+   </abaci>
+```
+
+```k
+    syntax MCDContract ::= AbacusContract
+    syntax AbacusContract ::= "Abacus" String
+    syntax MCDStep ::= AbacusContract "." AbacusStep [klabel(abacusStep)]
+ // ---------------------------------------------------------------------
+    rule contract(Abacus ABACUS_ADDRESS . _) => Abacus ABACUS_ADDRESS
+```
+
+### Constructor
+
+```k
+    syntax AbacusStep ::= "constructor" String
+ // ------------------------------------------
+    rule <k> Abacus ABACUS_ADDRESS . constructor _ ... </k>
+      (<abacus> <abacus-address> ABACUS_ADDRESS </abacus-address> ... </abacus> => .Bag )
+
+    rule <k> Abacus ABACUS_ADDRESS . constructor ABACUS_TYPE => . ... </k>
+         <msg-sender> MSGSENDER </msg-sender>
+         ( .Bag
+        => <abacus>
+            <abacus-address> ABACUS_ADDRESS     </abacus-address>
+            <abacus-type>    ABACUS_TYPE        </abacus-type>
+            <abacus-wards>   SetItem(MSGSENDER) </abacus-wards>
+            ...
+           </abacus>
+         )
+         [owise]
+```
+
+Abacus Authorization
+------------------
+
+```k
+    syntax AbacusStep ::= AbacusAuthStep
+    syntax AuthStep   ::= AbacusContract "." AbacusAuthStep [klabel(abacusStep)]
+ // ----------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


abaci.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+File-able Fields
+----------------
+
+These parameters are controlled by governance:
+
+-   `tau`:  Seconds after auction start when the price reaches zeroe
+-   `step`: Length of time between price drops
+-   `cut`:  Per-step multiplicative facto
```
kmbarry1, line 88:
```suggestion
-   `cut`:  Per-step multiplicative factor
```

abaci.md
```diff
@@ -0,0 +1,158 @@
...
+   syntax AbacusStep ::= "price" Ray Int
+ // -------------------------------------
+   rule <k> Abacus ABACUS_ADDRESS . price TOP DUR => ( TOP *Ray ray( ( ABACUS_TAU -Int DUR ) /Int ABACUS_TAU ) ) ... </k>
+      <abacus>
+         <abacus-address> ABACUS_ADDRESS   </abacus-address>
+         <abacus-type>    "LinearDecrease" </abacus-type>
+         <abacus-tau>     ABACUS_TAU       </abacus-tau>
+         ...
+      </abacus>
+      requires DUR >=Int ABACUS_TAU
```
kmbarry1, line 135: I think this condition should actually be `<=Int`. When `DUR > ABACUS_TAU`, the correct behavior is to evaluate to zero. Of course we don't use the linear decrease in practice anyway, so you could arguably omit it.

clip.md
```diff
@@ -0,0 +1,443 @@
...
+-----------------
+
+```k
+   syntax ClipStop ::= "noBreaker"
+                     | "noNewKick"
+                     | "noNewKickOrRedo"
+                     | "noNewKickOrRedoOrTake"
+
+   syntax Bool ::= ClipStop "<ClipStop" ClipStop [function]
+ // --------------------------------------------
```
kmbarry1, line 27: general nit: these separators don't match the length of the line above them in a lot of cases (Everett will complain about this, I know from experience)

clip.md
```diff
@@ -0,0 +1,443 @@
...
+    rule <k> emitTake ID MAX PRICE OWE TAB LOT USR => ID ... </k>
+         <return-value> ID:Int </return-value>
+         <frame-events> ... (.List => ListItem(Take(ID, MAX, PRICE, OWE, TAB, LOT, USR))) </frame-events>
+```
+
+
+
+
+
+
```
kmbarry1, line 178: general nit: could cut down on whitespace in various places

clip.md
```diff
@@ -0,0 +1,443 @@
...
+module CLIP
+    import DOG
+    import PRE-CLIP
+```
+
+```k
+   syntax ClipStep ::= "upchost"
+// -------------------------
+   rule <k> Clip . upchost => . ... </k>
+      <clip-chost> _ => DUST *Rad CHOP </clip-chost>
```
kmbarry1, line 302: `CHOP` is a `wad`...I don't think the correct units will result here, because IIUC `*Rad` assumes both arguments are `rads`.

kmbarry1: (I could be wrong, but it should be checked.)

clip.md
```diff
@@ -0,0 +1,443 @@
...
+   <clip-cusp> CLIP_CUSP </clip-cusp>
+   <clip-ilk> CLIP_ILK </clip-ilk>
+   <clip-tail> CLIP_TAIL </clip-tail>
+   <clip-tip> CLIP_TIP </clip-tip>
+   <clip-sales> ... ID |-> ClipSale( ... tab: TAB, lot: LOT, usr: USR, tic: (TIC => NOW),top: (TOP => ( ( Wad2Ray(VALUE) /Ray SPOT_PAR ) *Ray CLIP_BUF ) ) ) ... </clip-sales>
+   <clip-vat> CLIP_VAT </clip-vat>
+   <clip-vow> CLIP_VOW </clip-vow>
+   <current-time> NOW </current-time>
+   <spot-ilks> CLIP_ILK |-> SpotIlk(... pip: VALUE ) </spot-ilks>
+   <spot-par> SPOT_PAR </spot-par>
```
kmbarry1, line 338: consistency: we used just `PAR` above but here are using `SPOT_PAR`...

clip.md
```diff
@@ -0,0 +1,443 @@
...
+   => #fun(TOP_FINAL
+   => #fun(COIN
+   => ( #if ( CLIP_TIP >Rad rad(0) orBool CLIP_CHIP >Wad wad(0) ) #then ( #if ( TAB >=Rad CLIP_CHOST andBool ( Wad2Ray(LOT) *Ray ( Wad2Ray(VALUE) /Ray SPOT_PAR ) ) >=Ray CLIP_CHOST ) #then call CLIP_VAT . suck CLIP_VOW KPR COIN #else . #fi ) #else . #fi )
+   ~> DONE
+   ~> emitRedo ID TOP_FINAL TAB LOT USR KPR COIN
+   ) ( #if ( CLIP_TIP >Rad rad(0) orBool CLIP_CHIP >Wad wad(0) )  #then ( #if ( TAB >=Rad CLIP_CHOST andBool ( Wad2Ray(LOT) *Ray ( Wad2Ray(VALUE) /Ray SPOT_PAR ) ) >=Ray CLIP_CHOST ) #then CLIP_TIP +Rad ( Rad2Wad(TAB) *Wad CLIP_CHIP ) #else wad(0) #fi ) #else wad(0) #fi )
+   )( ( Wad2Ray(VALUE) /Ray SPOT_PAR ) *Ray CLIP_BUF )
+   )( ( (NOW -Int TIC ) >Int CLIP_TAIL ) orBool ( (PRICE /Ray TOP) <Ray CLIP_CUSP ) )
+   )( Abacus CLIP_ABACUS . price TOP (NOW -Int TIC) )
+   ... </k>
```
kmbarry1, line 324: I find this pretty hard to read/follow, is there a way to maybe indent it or otherwise format it to make it clearer? Or is this just the standard way to do thing in K and I need to get used to it :joy: ?

clip.md
```diff
@@ -0,0 +1,443 @@
...
+   <clip-stopped> CLIP_STOPPED </clip-stopped>
+   requires CLIP_STOPPED <ClipStop noNewKickOrRedo
+   andBool USR =/=K 0:Address
+   andBool ( ( Wad2Ray(VALUE) /Ray SPOT_PAR ) *Ray CLIP_BUF ) >Ray ray(0) //top > 0
+
+   rule <k> DONE:Bool ~> emitRedo ID TOP_FINAL TAB LOT USR KPR COIN => emitRedo ID TOP_FINAL TAB LOT USR KPR COIN ... </k>
+      requires DONE
+```
+
+TODO add external call
```
kmbarry1, line 348: Will this be added to this PR, or is it being saved for later?

clip.md
```diff
@@ -0,0 +1,443 @@
...
+    <this>       THIS       </this>
+    <msg-sender> MSG_SENDER </msg-sender>
+    requires CLIP_STOPPED <ClipStop noNewKickOrRedoOrTake
+    andBool USR =/=K 0:Address
+
+    rule <k> UPDATE:Bool ~> TAB:Rad ~> SLICE_INITIAL:Wad ~> OWE_INITIAL:Wad ~> PRICE:Ray ~> DONE:Bool ~> emitTake ID MAX PRICE OWE_FINAL TAB_FINAL LOT_FINAL USR => emitTake ID MAX PRICE OWE_FINAL TAB_FINAL LOT_FINAL USR ... </k>
+    <clip-sales> ... ID |-> ClipSale( ... tab: (TAB => #if UPDATE #then TAB_FINAL #else TAB #fi), lot:(LOT => #if UPDATE #then LOT_FINAL #else LOT #fi) ) ... </clip-sales>
+    <clip-chost> CLIP_CHOST </clip-chost>
+    requires DONE
+    andBool MAX >=Ray PRICE
```
kmbarry1, line 405: Not sure I fully understand what's going on here, but if `DONE` is true, the function reverts, whereas `MAX >=Ray PRICE` is a non-reversion condition...I suspect you want to negate `DONE` in this context.

abaci.md
0xverif: Checked on abaci.sol, fixed the condition and added a second rule for when it evaluates to 0

clip.md
0xverif: formatting errors fixed

clip.md
0xverif: formatting errors fixed

clip.md
0xverif: Checked and fixed, you were right, I didn't notice it which is weird because the compiler didn't complain, added a Wad2Rad conversion to CHOP.

clip.md
0xverif: I agree that it gets hard to follow fast but imo this method of declaring local variables is the cleanest/most understandable way of making a big function such as Clip's take readable. We can go over it in the meeting so you'll get a clearer understanding of it 🙌

clip.md
0xverif: changed to `SPOT_PAR` for consistency

clip.md
0xverif: It will be added to this PR but at the time I was unsure of how to model an external call in this framework, will talk with Everett about this

clip.md
0xverif: That it correct, added `notBool` before `DONE` and now the condition is negated.

dog.md
```diff
@@ -0,0 +1,204 @@
...
+   <vat-ilks> ... ILK |-> Ilk( ... rate: RATE, spot: SPOT, dust: DUST ) ... </vat-ilks>
+   <dog-dirt> DOG_DIRT </dog-dirt>
+   <dog-hole> DOG_HOLE </dog-hole>
+   requires ( ( SPOT >Ray ray(0) ) andBool ( (Wad2Ray(INK) *Ray SPOT ) <Ray (Wad2Ray(ART) *Ray RATE) ) )
+   andBool ( (DOG_HOLE >Rad DOG_DIRT ) andBool ( HOLE >Rad DIRT ) )
+
+
+   rule <k> ID:Int ~> emitBark ILK URN DINK DART DUE CLIP _ => emitBark ILK URN DINK DART DUE CLIP ID ... </k>
+   <vat-urns> ... { ILK, URN } |-> Urn( _, ART) ... </vat-urns>
+   <vat-ilks> ... ILK |-> Ilk( ... rate: RATE, spot: _, dust: DUST) ... </vat-ilks>
```
kmbarry1, line 178: Looks like some updates are missing here--`DOG_DIRT` needs to increase by `TAB`, as does the ilk-specific `DIRT`. Or maybe those belong in the rule above, but either way I don't them happening the way I see updates happening e.g. for `take`.

dog.md
```diff
@@ -0,0 +1,204 @@
...
+
+```k
+    configuration
+      <dog-state>
+         <dog-vat>   0:Address </dog-vat>
+         <dog-ilks>  .Map      </dog-ilks>   // mapping (bytes32 => Ilk) String |-> DogIlk
+         <dog-vow>   0:Address </dog-vow>
+         <dog-live>  true      </dog-live>
+         <dog-hole>  rad(1)    </dog-hole>
+         <dog-dirt>  rad(1)    </dog-dirt>
```
kmbarry1, line 25: On initial values--to match the contract behavior, these should start at zero. It doesn't matter much for `dog-hole` which is a governance parameter, but it actually is important for `dog-dirt` which needs to start at zero.

dog.md
```diff
@@ -0,0 +1,204 @@
...
+   <dog-dirt> DOG_DIRT </dog-dirt>
+   <dog-hole> DOG_HOLE </dog-hole>
+   requires ( ( SPOT >Ray ray(0) ) andBool ( (Wad2Ray(INK) *Ray SPOT ) <Ray (Wad2Ray(ART) *Ray RATE) ) )
+   andBool ( (DOG_HOLE >Rad DOG_DIRT ) andBool ( HOLE >Rad DIRT ) )
+
+
+   rule <k> ID:Int ~> emitBark ILK URN DINK DART DUE CLIP _ => emitBark ILK URN DINK DART DUE CLIP ID ... </k>
+   <vat-urns> ... { ILK, URN } |-> Urn( _, ART) ... </vat-urns>
+   <vat-ilks> ... ILK |-> Ilk( ... rate: RATE, spot: _, dust: DUST) ... </vat-ilks>
+   requires (#if (ART >Wad DART) #then (#if ( ( ART -Wad DART ) *Wad Ray2Wad(RATE) ) <Wad Rad2Wad(DUST)  #then true #else ( ( DART *Wad Ray2Wad(RATE) ) >=Wad Rad2Wad(DUST) ) #fi ) #else true #fi)
```
kmbarry1, line 179: The precision loss here may cause refinement proof failures--e.g. the actual code will compare `DART * RATE` directly against `DUST` (so `rad` vs `rad`)--converting down to `wad` here will lose information in some cases, leading to different results on certain groups of values between this spec and the actual contract behavior.

dog.md
```diff
@@ -0,0 +1,204 @@
...
+   => #fun(DART_FINAL
+   => #fun(DINK
+   => #fun(DUE
+   => #fun(TAB
+   => call DOG_VAT . grab ILK URN CLIP DOG_VOW (wad(0) -Wad DINK) (wad(0) -Wad DART_FINAL)
+   ~> call DOG_VOW . fess DUE
+   ~> call CLIP . kick TAB DINK URN KPR
+   ~> emitBark ILK URN DINK DART_FINAL DUE CLIP 0   
+   )( DUE *Wad CHOP )
+   )( DART_FINAL *Wad Ray2Wad(RATE) )
```
kmbarry1, line 158: Similar comment here about precision loss potentially being a monkey wrench--`DUE` is a `rad` in the code and will carry extra precision.

dog.md
```diff
@@ -0,0 +1,204 @@
...
+   ... </k>
+   <dog-vow> DOG_VOW </dog-vow>
+   <dog-live> true </dog-live>
+   <dog-vat> DOG_VAT </dog-vat>
+   <vat-urns> ... { ILK, URN } |-> Urn( INK, ART) ... </vat-urns>
+   <dog-ilks> ... ILK |-> Ilk( ... clip: CLIP:ClipContract, chop: CHOP, hole: HOLE, dirt: DIRT ) ...</dog-ilks>
+   <vat-ilks> ... ILK |-> Ilk( ... rate: RATE, spot: SPOT, dust: DUST ) ... </vat-ilks>
+   <dog-dirt> DOG_DIRT </dog-dirt>
+   <dog-hole> DOG_HOLE </dog-hole>
+   requires ( ( SPOT >Ray ray(0) ) andBool ( (Wad2Ray(INK) *Ray SPOT ) <Ray (Wad2Ray(ART) *Ray RATE) ) )
```
kmbarry1, line 172: And here `INK * SPOT` and `ART * RATE` would both be `rad`-precision and be compared against each other. Rounding will make the behavior different in some edge cases.

dog.md
```diff
@@ -0,0 +1,208 @@
...
+         <dog-ilks> ... ILK |-> Ilk( ... clip: (_ => CLIP) ) ... </dog-ilks>
+         <clip-ilk> CLIP_ILK </clip-ilk>
+      requires ILK ==K CLIP_ILK
+```
+
+Dog Semantics
+-------------
+
+```k
+    syntax DogStep ::= "bark" String Address Address
```
kmbarry1, line 147: Should probably keep ilk label typing consistent--it's a `String` here, but in the `dog-ilks` mapping, it's a `bytes32` (and `bytes32` always in the contracts).

dog.md
```diff
@@ -0,0 +1,208 @@
...
+        #let DART_INITIAL = minWad( ART, ( ( ROOM /Rate RATE) /Wad CHOP) ) #in (
+        #let DART_FINAL = (#if (ART >Wad DART_INITIAL) andBool ( ((ART -Wad DART_INITIAL) *Rate RATE) <Rad DUST ) #then ART #else   DART_INITIAL #fi) #in (
+        #let NON_DUSTY = (#if (ART >Wad DART_INITIAL) andBool ( ((ART -Wad DART_INITIAL) *Rate RATE) <Rad DUST ) #then true #else (DART_INITIAL *Rate RATE) >=Rad DUST  #fi) #in (
+        #let DINK = (INK *Wad DART_FINAL) /Wad ART #in (
+        #let DUE = DART_FINAL *Rate RATE #in (
+        #let TAB = DUE *RadWad2Rad CHOP #in (
+    NON_DUSTY
+    ~> KPR
+    ~> TAB
+    ~> emitBark ILK URN DINK DART_FINAL DUE CLIP 0 ) ) ) ) ) ) )
```
kmbarry1: Shouldn't the last argument of `emitBark` here be the auction id, not a hardcoded `0`?

dog.md
```diff
@@ -0,0 +1,208 @@
...
+```k
+    syntax DogAuthStep ::= "file" DogFile
+ // -------------------------------------
+
+    syntax DogFile ::= "vow"  Address
+                     | "Hole" Rad
+                     | "hole" String Rad
+                     | "chop" String Wad
+                     | "clip" String Address
+ // -----------------------------------------
```
kmbarry1:
```suggestion
 // ----------------------------------------
```

dog.md
```diff
@@ -0,0 +1,208 @@
...
+         <dog-dirt> DOG_DIRT  </dog-dirt>
+         <dog-hole> DOG_HOLE  </dog-hole>
+         <vat-urns> ... { ILK, URN } |-> Urn( INK, ART ) ...                                                </vat-urns>
+         <dog-ilks> ... ILK |-> Ilk( ... clip: CLIP:ClipContract, chop: CHOP, hole: HOLE,  dirt: DIRT ) ... </dog-ilks>
+         <vat-ilks> ... ILK |-> Ilk( ... rate: RATE, spot: SPOT, dust: DUST ) ...                           </vat-ilks>
+      requires DOG_LIVE ==Bool true
+      andBool ( ( SPOT >Ray ray(0) )
+      andBool ( (INK *Rate SPOT ) <Rad (ART *Rate RATE) ) )
+      andBool ( (DOG_HOLE >Rad DOG_DIRT )
+      andBool ( HOLE >Rad DIRT ) )
```
kmbarry1, line 172: I think we need `NON_DUSTY ==Bool true` as a side condition here as well.

kmbarry1: Oh maybe not, I see there's more below.

dog.md
```diff
@@ -0,0 +1,208 @@
...
+
+    rule <k> NON_DUSTY:Bool ~> KPR:Address ~> TAB:Rad ~> emitBark ILK URN DINK DART_FINAL DUE CLIP _
+    => call DOG_VAT . grab ILK URN CLIP DOG_VOW (wad(0) -Wad DINK) (wad(0) -Wad DART_FINAL)
+    ~> call DOG_VOW . fess DUE
+    ~> call CLIP . kick TAB DINK URN KPR
+    ~> emitBark ILK URN DINK DART_FINAL DUE CLIP 0
```
kmbarry1: Same comment here about the last event argument.

dog.md
0xverif: In the configuration declaration the `dog-ilks` mapping mentions in the comment next to it that it is a mapping from `(bytes32 => Ilk)` in solidity, but it is actually a `String |-> DogIlk` in this K spec, where the custom record `DogIlk` is declared in the middle of the document. We use `String` when the type in solidity is `bytes32` for simplification purposes. Looking at the code again I should rename that variable to `ILK_ID` which is in fact a much clearer name, where `ILK` sounds like we should be passing the `DogIlk` record as argument. Also maybe I should make the comment more explicit, but I followed the conventions that are established throughout the codebase (take vat.md as an example).

dog.md
```diff
@@ -0,0 +1,208 @@
...
+    ... </k>
+         <dog-vow>  DOG_VOW   </dog-vow>
+         <dog-vat>  DOG_VAT   </dog-vat>
+         <dog-ilks> ... ILK |-> Ilk( ... clip: CLIP:ClipContract ) ... </dog-ilks>
+         <vat-urns> ... { ILK, URN } |-> Urn( _, ART) ...                     </vat-urns>
+         <vat-ilks> ... ILK |-> Ilk( ... rate: RATE, spot: _, dust: DUST) ... </vat-ilks>
+      requires (#if (ART >Wad DART_FINAL) #then (#if ( ( ART -Wad DART_FINAL ) *Rate RATE ) <Rad DUST #then true #else ( ( DART_FINAL *Rate RATE ) >=Rad DUST ) #fi ) #else true #fi)
```
kmbarry1: This condition doesn't make sense to me. Let's take just the `true` branch of the first conditional:

```
#if (ART >Wad DART_FINAL) #then (#if ( ( ART -Wad DART_FINAL ) *Rate RATE ) <Rad DUST #then true #else ( ( DART_FINAL *Rate RATE ) >=Rad DUST ) #fi )
```

That's saying "if we did a partial liquidation, then if the remnant was was dusty, evaluate to `true`, else return the truth value of whether the auction itself is non-dusty". But we never leave a dusty remnant in the contract code--it always switches to a full a liquidation. And this logic seems to be encoded correctly above when determining `DART_FINAL`. So we only ever take one branch of that conditional. I'm not 100% sure but it looks like you're maybe just trying to re-express `NON_DUSTY` here and can drop the condition? Or am I missing what this actually captures from the code?

kmcd-data.md
```diff
@@ -69,9 +69,11 @@ We model everything with arbitrary precision rationals, but use sort information
  // -----------------------------------------
     rule Rad2Wad(FInt(R, RAD)) => FInt(R /Int RAY, WAD)
 
-    syntax Ray ::= Wad2Ray ( Wad ) [function]
+    syntax Ray ::= Wad2Ray ( Wad )      [function]
+                 | Wad2Ray ( MaybeWad ) [function]
```
kmbarry1: What do we need this `MaybeWad` variant for?

dog.md
0xverif: Edit: The formatting is p bad using the code blocks so I'm removing the `emitBark` definition and instead tell you to look at the file `dog.md`

The `BARK_ID` that should be emitted, and also be the `return-value`, in solidity is defined as such in the `bark` function ( a call to the clipper contract ):

`            id = ClipperLike(milk.clip).kick({
                tab: tab,
                lot: dink,
                usr: urn,
                kpr: kpr
            });
        }`
        
This is kind of a long explanation but I'll try to keep it concise 😅:

The 1st half of the K `dog.bark` definition (the one right below the syntax declaration) what we're doing is calculate every variable that is used throughout the function with the `#let` bindings + define some `requires` clauses that can be checked right away. Since with either the `#let` or the previously used `#fun` bindings we cannot use the temporary binded variables in other places of the definition, only on the cell they're defined, in the case the `<k>` cell, we use a _trick_ to be able to use them. By using the sequencing operator with all the needed temporary variables in order and adding the `emitBark` event in the end we can then create the 2nd half of the rule which will pattern match with this sequencing of variables and event, making the temp variables available for free use in this 2nd half. That's the reason why the 1st half doesn't update any state, make any function calls or use the temp variables for `requires` clauses. In the 2nd half we then do all that could not be done with the temporary variables in the 1st half. The reason why there's a `0` at the `emitBark ILK URN DINK DART_FINAL DUE CLIP 0` is pretty much the same. There's no way to _save_ the result  of the call to `clip` in K and then re-use it when calling the emitBark event. It would require something like calling it again in the `emitBark` call and then let the computation unfold until there's the integer we need, which would be incorrect. That's why `emitBark` is defined as such:

(Check the Dog Events section, in particular the emitBark rule)

This will pattern match to the end of the 2nd half of the `bark` rule, which right before the call to `emitBark` will have the return of the `clip` call that we need, and then correctly emit the event and place the proper `return-value`. 

I can go over this in the next week meeting as well.

kmcd-data.md
0xverif: We don't, my bad, removing it

dog.md
0xverif: Not sure if I can resolve this one?

0xverif: I think I captured all those `requires` clauses

dog.md
kmbarry1: Okay gotcha. Yeah, `ILK_ID` probably a little clearer although like you say, not inconsistent.

dog.md
kmbarry1: I'm assuming the below is some sort of continuation of processing the `bark` logic so there's need to require `NON_DUSTY` here. If that's at least roughly correct go ahead and resolve.

0xverif: Yep yep, further explained in the big answer of the `emitBark` event

dog.md
kmbarry1: Thanks for the explanation! I think I would need a better understanding of K to really grasp this, but at my current level it basically makes sense. Would it be correct to say that the `emitBark` expression here doesn't emit anything yet, it's just needed to pattern match with the second half of the `bark` definition?

dog.md
kmbarry1: Okay, given the discussion above I'd expect to see here some non-zero value for the id, even if we cheat and read it from the Clipper's storage. Or maybe I'm still missing something :sweat_smile: .

clip.md
```diff
@@ -0,0 +1,470 @@
...
+ // --------------------------------------------------------------------------------------------------------------------------------
+
+    syntax ClipStep ::= "emitKick" Int Ray Rad Wad Address Address Wad
+ // ------------------------------------------------------------------
+    rule <k> emitKick KICK_ID TOP TAB LOT USR KPR COIN => KICK_ID ... </k>
+         <return-value> KICK_ID:Int </return-value>
+         <frame-events> ... (.List => ListItem(Kick(KICK_ID, TOP, TAB, LOT, USR, KPR, COIN))) </frame-events>
+
+    syntax CustomEvent ::= Redo(id: Int, top: Ray, tab: Rad, lot: Wad, usr: Address, kpr: Address, coin: Wad) [klabel(Redo), symbol]
+ //---------------------------------------------------------------------------------------------------------------------------------
```
kmbarry1, line 159: formatting

clip.md
```diff
@@ -0,0 +1,470 @@
...
+
+```k
+    syntax ClipLockStep ::= "take" Int Wad Ray Address String
+ // ---------------------------------------------------------
+    rule <k> Clip . take TAKE_ID AMT MAX WHO DATA
+    =>  #let PRICE = Abacus CLIP_ABACUS . price TOP (NOW -Int TIC) #in (
+        #let DONE = ( (NOW -Int TIC ) >Int CLIP_TAIL ) orBool ( (PRICE /Ray TOP) <Ray CLIP_CUSP ) #in (
+        #let SLICE_INITIAL = minWad(LOT, AMT) #in (
+        #let OWE_INITIAL = SLICE_INITIAL *WadRay2Rad PRICE #in (
+        #let OWE_FINAL = ( #if (OWE_INITIAL >Rad TAB) #then TAB #else ( #if (OWE_INITIAL <Rad TAB andBool SLICE_INITIAL <Wad LOT ) #then (#if ( TAB -Rad OWE_INITIAL <Rad Ray2Rad(CLIP_CHOST) ) #then (TAB -Rad Ray2Rad(CLIP_CHOST) ) #else OWE_INITIAL #fi ) #else OWE_INITIAL #fi ) #fi ) #in (
```
kmbarry1: Hmmmm `CLIP_CHOST` should already be a `rad`, not sure why the `Ray2Rad` conversion is needed?

kmbarry1: (ditto for other occurrences of `CLIP_CHOST` in the `take` rules)

dog.md
0xverif: I am in fact, re-expressing `NON_DUSTY`, but incorrectly by using the final values of the mentioned variables, I'm removing this condition

0xverif: Thanks!

dog.md
0xverif: Yes that is correct, and it is also exactly the same case in the second part of the rule, we only want it to pattern match with the `emitBark` rule which will then properly emit the event and set the return value. Complementing this answer with the one on the next, related, comment.

dog.md
0xverif: I did not include `id` in the Clippers storage, I believe `id` is not a variable that is maintained in storage in the solidity implementation as well right? I believe it is only used in memory. It is still a 0 value and I explain why on the comment where I tagged you.

dog.md
```diff
@@ -0,0 +1,207 @@
...
+Dog Events
+----------
+
+```k
+    syntax CustomEvent ::= Bark(ilk: String, urn: Address, ink: Wad, art: Wad, due: FInt, clip: Address, Id: Int) [klabel(Bark), symbol]
+ //-------------------------------------------------------------------------------------------------------------------------------------
+
+    syntax DogStep ::= "emitBark" String Address Wad Wad FInt Address Int
+ // --------------------------------------------------------
+    rule <k> BARK_ID:Int ~> emitBark ILK URN INK ART DUE CLIP _ => BARK_ID ... </k>
```
0xverif: @kmbarry1  here on the rule definition of `emitBark` we tell K that we _do not care_ about the concrete value of the integer that comes after `CLIP`, the 0 on the other rules, and in fact the return value is the Integer that is behind the `emitBark`, the `BARK_ID` that returns from the `Clip.kick` call on the 2nd part of the `Bark` rule.

I understand that these circumventions are quite complex if one doesn't read/write K everyday, I used this trick to remain faithful to the solidity implementation of the `Bark `event interface, but I can just simply remove the unused Int from the syntax and semantics so that this doubt does not arise when reading this code, since I bet everyone reading this would be confused. Tell me what you consider best, thanks!

clip.md
0xverif: That is correct and I've also noticed a discrepancy in the `redo` rule. Fixing both

dog.md
kmbarry1: Coming back to this a few months later--maybe I finally understand. The auction id is going to come from the contents of the k cell prior to continuing through to `emitBark`, because it is the return value of `kick`, right?

dog.md
kmbarry1: Now that I finally (mostly) understand I'd personally favor removing the unused `Int`...unless that will interfere with refinement proofs in some way (don't think it should though).

kmbarry1: Could also add a short comment noting that even though the signature doesn't match, the emitted data will be the same.

clip.md
kmbarry1: Looks like the fixes didn't get pushed?

kmbarry1 (approved):
LGTM

0xverif:
> LGTM

Sorry for taking so long @kmbarry1! I formatted my pc since we spoken and so I lost the local fixes I had, the ones I pushed are now completely fixed and compiling

0xverif:
@ehildenb CI is failing with an OOM error, can you increase the memory allowed?

kmbarry1:
> > LGTM
> 
> Sorry for taking so long @kmbarry1! I formatted my pc since we spoken and so I lost the local fixes I had, the ones I pushed are now completely fixed and compiling

No worries! Guess I LGTMed a little prematurely (forgot about what all neeeded fixed), but seems good now.

kmbarry1:
Hmmmm merging seems to be blocked on the CI--is this something that should be fixed, or should we just override that requirement and merge on our side @0xverif (it seems this repo is using the RV Jenkins instance)?

0xverif:
Update: Waiting for help from the K frontend team.

